Issue #31, opened 2020-10-09
Improve Storybook Example with PayPal Marks radio buttons


gregjopa:
The [current radio buttons example](https://paypal.github.io/react-paypal-js/?path=/docs/example-paypalmarks--radio-buttons) works but could be improved to use the `args` and follow more React best practices. 

Please update this component to use this structure:

```js
return (
    <PayPalScriptProvider
        options={{
            "client-id": "sb",
            components: "buttons,marks,funding-eligibility",
        }}
    >
        {/* get the fundingSources list from args and dynamically create the list of radio buttons from it */}
        <FormWithRadioButtons {...args } onChange={onChange} />
        <PayPalButtons fundingSource={fundingSource} />
    </PayPalScriptProvider>
);
```

```js
export const RadioButtons = RadioButtonTemplate.bind({});
RadioButtons.args = { fundingSources: [FUNDING.PAYPAL, FUNDING.CARD, FUNDING.PAYLATER] };
```

PR #34, opened 2020-10-13
docs/paypal-marks-radio-buttons

PaulieScanlon:
Hey team, great job on this lib!

I've updated `PayPalMarks.stories.js` to dynamically create the list of radio buttons as mentioned in [#31]

However. Arguably Storybook becomes less useful when you abstract DOM arrangement / creation into functions. E.g `FormWithRadioButtons` does indeed `map` over `fundingSources` to dynamically create the list of radio buttons but when you visit the "Show code" snippet the end _user_ can't see how those radio buttons / Paypal Marks have been created as they can't see the props being used to create the Story:

E.g looking at the "Show code" all we see is the `FormWithRadioButtons` component.

```javascript
...
  <FormWithRadioButtons
    fundingSources={[
      'paypal',
      'card',
      'paylater'
    ]}
    onChange={() => {}}
  />
...
</PayPalScriptProvider>
```

What might be a better approach is to _not_ abstract `FormWithRadioButtons`, but instead use the `map` "inline in the story"

```javascript
function RadioButtonTemplate(args) {
...
        <form>
            {args.fundingSources.map((source, index) => (
                <label className="mark" key={index}>
                    <input
                        defaultChecked={index === 0 ? true : false}
                        onChange={onChange}
                        type="radio"
                        name="fundingSource"
                        value={source}
                    />
                    <PayPalMarks fundingSource={source} />
                </label>
            ))}
        </form>
...
}
```

Which results in a more representative "Show code" snippet. 

E.g

```javascript
...
    <label className="mark">
      <input
        defaultChecked
        name="fundingSource"
        onChange={() => {}}
        type="radio"
        value="paypal"
      />
      <Marks fundingSource="paypal" />
    </label>
...
```

I'm happy to make that change and push again if you're in agreement?

On that note, it is then a little confusing that the usage for `PayPalMarks` is `<PayPalMarks />` but when exported and rendered in the "Show code" snippet it displays as `<Marks />`

I believe that comes from here [index.js#L2](https://github.com/paypal/react-paypal-js/blob/main/src/index.js#L2) and here [PayPalMarks.js#L26](https://github.com/paypal/react-paypal-js/blob/main/src/components/PayPalMarks.js#L26)

I wonder perhaps if declaring this component as `PayPalMarks` rather than `Marks` would solve that problem?

That said, this was fun. I'd like to continue to contribute as i'm a bit of a component library mega dweeb and love all these pedantic problems...

And finally, **naming things is hard**

### DIFF
--- a/src/stories/PayPalMarks.stories.js
+++ b/src/stories/PayPalMarks.stories.js
@@ -62,7 +62,7 @@ function RadioButtonTemplate(args) {
                 {args.fundingSources.map((source, index) => (
                     <label className="mark" key={index}>
                         <input
-                            defaultChecked={index === 0 ? true : false}
+                            defaultChecked={index === 0}
                             onChange={onChange}
                             type="radio"
                             name="fundingSource"
--- a/src/stories/PayPalMarks.stories.js
+++ b/src/stories/PayPalMarks.stories.js
@@ -44,10 +44,31 @@ StandAlone.parameters = {
     },
 };
 
+const FormWithRadioButtons = (args) => {
+    const { fundingSources, onChange } = args;
+
+    return (
+        <form>
+            {fundingSources.map((source, index) => (
+                <label className="mark" key={index}>
+                    <input
+                        defaultChecked={index === 0 ? true : false}
+                        onChange={onChange}
+                        type="radio"
+                        name="fundingSource"
+                        value={source}
+                    />
+                    <PayPalMarks fundingSource={source} />
+                </label>
+            ))}
+        </form>
+    );
+};
+
 function RadioButtonTemplate(args) {
     const [fundingSource, setFundingSource] = useState(FUNDING.PAYPAL);
 
-    function onClick(event) {
+    function onChange(event) {
         setFundingSource(event.target.value);
     }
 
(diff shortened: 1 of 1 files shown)


src/stories/PayPalMarks.stories.js
```diff
@@ -44,10 +44,31 @@ StandAlone.parameters = {
...
 
+const FormWithRadioButtons = (args) => {
+    const { fundingSources, onChange } = args;
+
+    return (
+        <form>
+            {fundingSources.map((source, index) => (
+                <label className="mark" key={index}>
+                    <input
+                        defaultChecked={index === 0 ? true : false}
```
elizabethmv, line 55: Small request, this line could be simplified.
```suggestion
                        defaultChecked={index === 0}
```

elizabethmv (commented):
Hi @PaulieScanlon, thanks for your work on this!

We agree with the suggestion to remove the abstraction and use the inline approach with `map`. It does show a better code snippet. Please go ahead and push up those changes!

Additionally, would you be up for creating a new PR to change `Marks` to `PayPalMarks` ? As mentioned, changing the following lines should resolve the component name issue:
https://github.com/paypal/react-paypal-js/blob/main/src/components/PayPalMarks.js#L26
https://github.com/paypal/react-paypal-js/blob/main/src/components/PayPalMarks.js#L65

PaulieScanlon:
> Hi @PaulieScanlon, thanks for your work on this!
> 
> We agree with the suggestion to remove the abstraction and use the inline approach with `map`. It does show a better code snippet. Please go ahead and push up those changes!
> 
> Additionally, would you be up for creating a new PR to change `Marks` to `PayPalMarks` ? As mentioned, changing the following lines should resolve the component name issue:
> https://github.com/paypal/react-paypal-js/blob/main/src/components/PayPalMarks.js#L26
> https://github.com/paypal/react-paypal-js/blob/main/src/components/PayPalMarks.js#L65

Hey @elizabethmv absolutely fine to make those changes and sure thing, i'll create another PR to change marks. Ty for reviewing the code!

src/stories/PayPalMarks.stories.js
PaulieScanlon: Happy to make this change... but.... in the interest of surfacing useful code snippets is the ternary clearer ... we just don't know haha!

elizabethmv:
@PaulieScanlon could you rebase with the main branch? It looks like the CI build may be failing because your branch is missing the commit for Github Actions that was added earlier today.

PaulieScanlon:
@elizabethmv Sure thing... let's see if this works!

elizabethmv (approved):
Looks good! Thanks again for your contribution @PaulieScanlon! 🥇 🎉

PR #41, opened 2020-10-15
Add Github Action to deploy Storybook

elizabethmv:
## What
This PR adds a Github `storybook.yml` to deploy storybook on pushes/merges to main. It utilizes the [path workflow](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths) which allows us to check for any changes to `src/` or the README (synced with our Getting Started page). This way we only deploy on main when there are changes available.

This PR also adds an update to`main.yml` to check storybook builds on every pull request.

## Why
Why _not_ automate storybook deploys?

### DIFF
--- a/.github/workflows/storybook.yml
+++ b/.github/workflows/storybook.yml
@@ -1,0 +1,18 @@
+name: "Deploy Storybook to Github Pages"
+
+on:
+  push:
+    branches:
+      - main
+    paths: ["stories/**", "src/components/**"]
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+
+  steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Deploy Storybook
+        run: npm run deploy-storybook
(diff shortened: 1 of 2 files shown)


.github/workflows/storybook.yml
```diff
@@ -0,0 +1,23 @@
...
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run: npm install
+
+    - name: Deploy Storybook
+      run: npm run deploy-storybook -- --ci
+      env:
+        GH_TOKEN: ${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}
```
elizabethmv, line 23: Docs on [Actor](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions) and the [Token](https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow)

gregjopa (approved):
This is an awesome improvement! It's great to see GitHub Actions deploying storybook to github pages when PRs are merged. 💯 🥇

PR #43, opened 2020-10-16
Add 'resetOptions' action for dispatcher for reloading use case

gregjopa:
This PR adds a generic `resetOptions` action for changing the sdk params with the dispatcher. The dispatcher is used for the reload use case. For example, the following use cases would use `resetOptions`:
- Updating currency
- Toggling the transaction type from Order to Subscription

Originally I added the `updateCurrency` action thinking each use case could have it's own specific action. Looking back, I think that was a mistake because there are so many different combinations of params that could be sent to the sdk, and we can't add a specific action for every use case. Instead, let's have one function that's able to set params and pass them through to paypal-js to reload the sdk. We can remove `updateCurrency` when we do a major version bump.

### DIFF
--- a/src/ScriptContext.js
+++ b/src/ScriptContext.js
@@ -14,13 +14,8 @@ function scriptReducer(state, action) {
                 },
                 isLoaded: action.value,
             };
-        case "updateOptions":
-            return {
-                options: {
-                    ...state.options,
-                    ...action.value,
-                },
-            };
+        case "resetOptions":
+            return { options: action.value };
 
         // deprecated - remove for the v2 release
         case "changeCurrency":
--- a/src/ScriptContext.js
+++ b/src/ScriptContext.js
@@ -14,6 +14,15 @@ function scriptReducer(state, action) {
                 },
                 isLoaded: action.value,
             };
+        case "updateOptions":
+            return {
+                options: {
+                    ...state.options,
+                    ...action.value,
+                },
+            };
+
+        // deprecated - remove for the v2 release
         case "changeCurrency":
             return {
                 options: {
--- a/src/stories/Subscriptions.stories.js
+++ b/src/stories/Subscriptions.stories.js
@@ -15,17 +15,13 @@ export default {
 const scriptOptions = {
     "client-id": "sb",
     components: "buttons",
+    intent: "subscription",
+    vault: true,
 };
 
 function Template(args) {
-    const options = {
-        ...scriptOptions,
-        intent: "subscription",
-        vault: true,
-    };
-
     return (
-        <PayPalScriptProvider options={options}>
+        <PayPalScriptProvider options={scriptOptions}>
             <PayPalButtons {...args} />
         </PayPalScriptProvider>
     );
@@ -60,11 +56,7 @@ function TransactionTypeForm() {
         if (event.target.value === "subscription") {
             dispatch({
                 type: "resetOptions",
-                value: {
-                    ...scriptOptions,
-                    intent: "subscription",
-                    vault: true,
-                },
+                value: scriptOptions,
             });
         } else {
             dispatch({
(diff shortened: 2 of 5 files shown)


src/ScriptContext.js
```diff
@@ -14,6 +14,15 @@ function scriptReducer(state, action) {
                 },
                 isLoaded: action.value,
             };
+        case "updateOptions":
+            return {
+                options: {
```
gregjopa, line 19: I thinking we should add additional logic here to strip out any keys with invalid values. Right now, there's no way to remove a query parameter key after it's been added. The orders/subscriptions demo works because the query params change like so:
- ?intent=subscription&vault=true
- ?intent=order&vault=false

However, it would not work if we needed to do this since there's no way to currently drop a key once it's been added:
- ?intent=subscription&vault=true
- ?intent=order  

Also, note that the following approach can't be used since the js sdk validates query param values to ensure they are valid: https://www.paypal.com/sdk/js?client-id=sb&intent=order&vault= 



I'm thinking that passing `""`, `null`, or `undefined` for a value would cause a key to be dropped. For example, this would drop vault:
```js
dispatch({ action: "updateOptions", value: { intent: "order", vault: "" }
```

Thoughts?

mnicpt: @gregjopa I didn't see where we were validating valid keys.  Are you planning on adding that validation?  Also, removing an option is an action and think we should have a `removeOption` action instead of having the customer remove by value.

gregjopa: @mnicpt I agree that `removeOption` would see more conventional for the api. But I'm concerned it would introduce multiple actions needing to be called when reloading. For example, here's what it would look like when switching from Subscription to Order:

```js
dispatch({ action: "updateOptions", value: { intent: "order" })
dispatch({ action: "removeOption", value: 'vault' })
```

With the current architecture, this will trigger two reloads of the script. We could technically write logic to batch up all these dispatched actions in x milliseconds but that seems like a problem we could avoid all together if we continue to follow the design of one dispatched action equaling one network request to reload the JS SDK. 

Based on this I'd prefer these `actions`:
- `updateOptions` - update the existing query params (perhaps this shouldn't allow for dropping of query params, only changing values 🤔) 
- `newOptions` for dropping all the existing query params and starting fresh with defining all new query params. I haven't coded this action yet but I think this could be the solution for folks looking to make several changes to query params and not just updating values for a couple keys.

mnicpt: @gregjopa See what you mean.  Initially, I was thinking that `updateOptions` overwrote the prior ones as that was more indicative of what would happen if you reloaded the script.  What do you think of that?  Only overwriting and not merging options so there is only 1 action. The customer will initially define the options as an object and even could use the `useState` hook on these so when that state is updated, they could dispatch the `updateOptions` action with the new state.

gregjopa: @mnicpt I like that idea! Nice suggestion 😄 It keeps our `action` implementation simple and provides developers with the flexibility to change any thing with the query string.

I'll take a stab at updating the PR to use this approach for the Change Currency and Orders vs Subscriptions demos. Perhaps we name the action `setOptions` instead of `updateOptions` to make it clear that it is resetting the query string to the new values for the reload use case.

mnicpt: Or even `resetOptions`. 🧐

gregjopa: I like `resetOptions`! I went with that. Just pushed an update. 

I'm going to add some unit testing and this should be ready to go.

mnicpt (approved):
Awesome, @gregjopa !!!

src/stories/Subscriptions.stories.js
```diff
@@ -34,3 +38,82 @@ Default.args = {
...
+                },
+            });
+        }
+    }
+
+    return (
+        <form>
+            <label>
+                <input
+                    defaultChecked
```
gregjopa, line 77: @elizabethmv pointed out a bug with the Subscriptions storybook docs page. On this story, the demos need to share the same JS SDK script and defaulting to Order in the Toggle demo was breaking the first demo. To resolve the issue, I updated it to default to Subscriptions for the toggle demo so now both demos work in the Docs view by default:

Screen Shot 2020-10-21 at 1 34 19 PM

The second demo will still break the first demo when toggling to Orders. But that's due to the fact that you cannot have a Order button and Subscription button on the same page together with the JS SDK.

elizabethmv (approved):

PR #48, opened 2020-10-25
Track script loading error

### LINKED ISSUE #46: [Feature] Expose script error state
### 🚀 Feature Proposal

The library should expose the failing state of the SDK script

### Motivation

There's currently no indication in case the SDK script fails to load for any reason so there's no way for the application to figure out that the script has failed

### Example

This information would be useful to have the ability to notify the app user that they should take some actions like reloading the page or checking their network connection

if the script fails to load loadScript [rejects](https://github.com/paypal/paypal-js/blob/cbe0bf00344e3d105e9d66e7a6a618b757842a3c/src/main.js#L32-L34) but 
PayPalScriptProvider does not catch the rejected Promise
https://github.com/paypal/react-paypal-js/blob/d0a0f38881fc73abafac5302eb2873882719f328/src/ScriptContext.js#L59-L62

We can replace `isLoaded` state with a `loading/loaded/error` state. This would be a breaking change so we can first introduce a new field while keeping `isLoaded` for the backward compatibility

If you're interested I can take a stab at this

dreyks:
closes #46 

So... I've made the change but to test it I had to rewrite tests a bit and mock the `loadScript` instead of testing the real `script` tag. That lead to changes in buttons and markers test and I got a bit carried away :D 

Feel free to request any reverts :)


src/components/PayPalButtons.js
```diff
@@ -59,7 +60,10 @@ function verifyGlobalStateForButtons({ components = "" }) {
             "\nTo fix the issue, add 'buttons' to the list of components passed to the parent PayPalScriptProvider:" +
             `\n\`<PayPalScriptProvider options={{ components: '${expectedComponents}'}}>\`.`;
     }
-    throw new Error(errorMessage);
+    setErrorState(() => {
+        throw new Error(errorMessage);
+    });
```
dreyks, line 65: Legalized hack by Dan to be able to catch async errors with ErrorBoundary
https://github.com/facebook/react/issues/14981#issuecomment-468460187

src/components/PayPalButtons.js
```diff
@@ -12,18 +12,19 @@ import { usePayPalScriptReducer } from "../ScriptContext";
...
-    const [{ isLoaded, options }] = usePayPalScriptReducer();
+    const [{ loadingStatus, options }] = usePayPalScriptReducer();
     const buttonsContainerRef = useRef(null);
     const buttons = useRef(null);
+    const [, setErrorState] = useState(null);
 
     useEffect(() => {
-        if (isLoaded) {
-            verifyGlobalStateForButtons(options);
+        if (loadingStatus === "resolved") {
```
gregjopa, line 21: For readability, could you flatten out these conditions? Perhaps something like:

```js
if (loadingStatus !== "resolved") {
    return;
}

if (!hasValidGlobalState(options, setErrorState)) {
    return;
}

if (!buttons.current.isEligible()) {
    return;
}

// and finally render the buttons
```

src/components/PayPalButtons.js
```diff
@@ -42,9 +43,9 @@ export default function PayPalButtons(props) {
     return <div ref={buttonsContainerRef} />;
 }
 
-function verifyGlobalStateForButtons({ components = "" }) {
+function verifyGlobalStateForButtons({ components = "" }, setErrorState) {
```
gregjopa, line 46: Now this this returns a boolean, can we rename it to something that begins with `is` or `has`? Something like:
```suggestion
function hasValidGlobalState({ components = "" }, setErrorState) {
```

src/components/PayPalButtons.js
dreyks: I thought about that too, but we'll have to be cautious with the return value so that we don't skip the cleanup

for example in case `!button.current.isEligible()` we still better return the cleanup function so that the next effect returns a clean state

i was trying to make minimal changes but i can give it a try

src/components/PayPalButtons.test.js
```diff
@@ -38,44 +43,72 @@ describe("<PayPalButtons />", () => {
...
+        expect(onError.mock.calls[0][0].message).toMatchSnapshot();
     });
 });
 
-// immediately sets the PayPalScriptProvider `isLoaded` state to true
-function SetIsLoadedToTrue() {
-    const [, dispatch] = usePayPalScriptReducer();
-    dispatch({ type: "setIsLoaded", value: true });
-    return null;
+class ErrorBoundary extends React.Component {
```
gregjopa, line 95: This `ErrorBoundary` component is great! It seems like a convenient way to unit test async errors. And using it with the `wrapper` option with the `render()` test helper is a nice touch 💯

src/components/PayPalButtons.test.js
```diff
@@ -17,16 +24,14 @@ describe("<PayPalButtons />", () => {
     });
 
     test("should pass props to window.paypal.Buttons()", async () => {
-        window.paypal.Buttons = () => {
-            return {
+        window.paypal = {
```
gregjopa, line 27: Nice readability improvement with being able to use `window.paypal.Buttons` directly as a reference for the spy 👏

dreyks:
i'll make changes and rebase on top of the current main branch

src/components/PayPalMarks.js
```diff
@@ -24,28 +24,29 @@ import { usePayPalScriptReducer } from "../ScriptContext";
...
-    const [{ isLoaded, options }] = usePayPalScriptReducer();
+    const [{ loadingStatus, options }] = usePayPalScriptReducer();
     const markContainerRef = useRef(null);
     const mark = useRef(null);
+    const [, setErrorState] = useState(null);
 
     useEffect(() => {
-        if (isLoaded && !mark.current) {
-            verifyGlobalStateForMarks(options);
+        if (loadingStatus === "resolved" && !mark.current) {
```
gregjopa, line 33: Same comment here as with Buttons about flattening out the conditions.

src/components/PayPalMarks.js
```diff
@@ -24,28 +24,29 @@ import { usePayPalScriptReducer } from "../ScriptContext";
...
+                }
             }
         }
     });
 
     return <div ref={markContainerRef} />;
 }
 
-function verifyGlobalStateForMarks({ components = "" }) {
+function verifyGlobalStateForMarks({ components = "" }, setErrorState) {
```
gregjopa, line 47:
```suggestion
function hasValidGlobalState({ components = "" }, setErrorState) {
```

src/components/PayPalButtons.test.js
```diff
@@ -1,13 +1,20 @@
...
+import PropTypes from "prop-types";
 import { render, waitFor } from "@testing-library/react";
 
-import { PayPalScriptProvider, usePayPalScriptReducer } from "../ScriptContext";
+import { PayPalScriptProvider } from "../ScriptContext";
 import PayPalButtons from "./PayPalButtons";
 import { FUNDING } from "@paypal/sdk-constants";
+import { loadScript } from "@paypal/paypal-js";
+
+jest.mock("@paypal/paypal-js", () => ({
```
gregjopa, line 10: We do have a shared mock for paypal-js here: https://github.com/paypal/react-paypal-js/blob/main/__mocks__/%40paypal/paypal-js.js. Is there a problem with using it? Happy to tweak it. Ideally we'd use it with the default auto-mock setting so components like Buttons, Marks, etc. don't need to worry about mocking script loading.

dreyks: hmm... never saw this

but in my opinion, it still kinda tests the internal implementation of `@paypal/paypal-js` which is already [covered with tests](https://github.com/paypal/paypal-js/blob/cbe0bf00344e3d105e9d66e7a6a618b757842a3c/src/main.test.js)

i think we can safely mock `loadScript` everywhere in this library

gregjopa:
Thanks for the PR @dreyks. This is great! I've learned a lot from it 💯

I pulled down your branch and ran through the storybook demos. There are a couple things I noticed:

1. Could you please update the LoadingSpinner story to work with this change? https://github.com/paypal/react-paypal-js/blob/main/src/stories/usePayPalScriptReducer.stories.js#L91

It needs to be updated to use the new `loadingStatus` state with something like:

```js
function LoadingIndicator() {
    const [{ loadingStatus }] = usePayPalScriptReducer();
    const isLoaded = loadingStatus === 'resolved';

    if (isLoaded) return null;
    if (!isLoaded) return <div className="spinner" />;
}

```

2. The `<PayPalMessages />` component needs to be updated to use `loadingStatus`: https://github.com/paypal/react-paypal-js/blob/main/src/components/PayPalMessages.js#L10

The rest of the stories look great! Happy to merge this after you make these minor changes.

src/components/PayPalButtons.js
gregjopa: Ah I forgot about needing to sometimes return the cleanup. Good point. Perhaps we can do something like:

```js
function cleanup() {
    // close the buttons when the component unmounts
    if (buttons.current) {
        buttons.current.close();
    }
}

if (loadingStatus !== "resolved") {
    return;
}

if (!hasValidGlobalState(options, setErrorState)) {
    return;
}

if (!buttons.current.isEligible()) {
    return cleanup;
}
```

dreyks: yup. that'd work

src/components/PayPalButtons.test.js
gregjopa: I'm sold! We can delete the paypal-js mock that's in this repo. Originally I thought we could use the DOM to verify the script is loaded but that adds unnecessary complexity. And you make a good point that paypal-js is already unit tested and we should keep those testing concerns separate.

src/components/PayPalButtons.js
```diff
@@ -12,45 +12,47 @@ import { usePayPalScriptReducer } from "../ScriptContext";
...
-        return () => {
-            // close the buttons when the component unmounts
+        const cleanup = () => {
             if (buttons.current) {
                 buttons.current.close();
             }
         };
+
+        if (loadingStatus !== "resolved") {
+            return cleanup;
```
dreyks, line 28: we have to return cleanup here too to support reloading the script

basically, we probably should always return the cleanup function just to be safe

gregjopa: I agree, we should always return the cleanup function. The new flat structure looks good! 😄

src/ScriptContext.js
```diff
@@ -24,7 +27,7 @@ function scriptReducer(state, action) {
                     ...state.options,
                     currency: action.value,
                 },
-                isLoaded: false,
+                loadingStatus: "pending",
```
gregjopa, line 30: As I review this for the second time I'm a little worried about these magic string values for `loadingStatus`. One pattern we use in our other PayPal repos is to have a constants.js file that defines these possible strings values like so:

```js
// src/constants.js
export const SCRIPT_LOADING_STATE = {
    PENDING: 'pending',
    REJECTED: 'rejected',
    RESOLVED: 'resolved',
};

```

Would you mind making this change? It may seem like overkill but I think it would be worth it. The thinking here is ScriptContext.js and all the components and the tests would import this constants file and use it like so:

```js
import { SCRIPT_LOADING_STATE } from 'src/constants.js';

// and then use it in the `useEffect` hook like so:
if (loadingStatus !== SCRIPT_LOADING_STATE.RESOLVED) {
   return cleanup;
}
```

We could also publicly export these constants [here](https://github.com/paypal/react-paypal-js/blob/main/src/index.js) so folks using this library could leverage them when building custom components based on state like the loading spinner example. 

Let me know your thoughts on this change.

gregjopa (approved):
Awesome work @dreyks! 💯 🥇 👍 

The new changes look great. I left one non-blocking comment about adding a `constants.js` file for the `loadingStatus` values.

These new state changes are awesome! 👏 For example, if I test things out with a disabled internet connection I can now detect that the script failed to load based on state exposed by `usePayPalScriptReducer`. Previously, this wasn't possible.

react-dev-tools-error-state-screenshot

dreyks:
yup. dealing with network connection issues is something that we're doing while loading paypal sdk manually and that was the reason i couldn't simply move to react-paypal-js

regarding named constants: that's a great idea. I actually wanted to introduce something like this but decided that it'd be too many changes for such a "simple" feature change (haha)

one alternative to constants is using derived functions. something like

```js
function usePayPalScriptReducer() {
    const scriptContext = useContext(ScriptContext);
    const dispatchContext = useContext(ScriptDispatchContext);
    if (scriptContext === undefined || dispatchContext === undefined) {
        throw new Error(
            "useScriptReducer must be used within a ScriptProvider"
        );
    }

    const { loadingStatus, ...restScriptContext } = scriptContext

    const derivedStatusContext = {
        ...restScriptContext,
        isPending: loadingStatus === 'pending',
        isResolved: loadingStatus === 'resolved',
        isRejected: loadingStatus === 'rejected'
    }

    return [derivedStatusContext, dispatchContext];
}
```

notice how I do not expose `loadingStatus` at all, it just serves as a source of truth for the derived properties. We can even still extract those strings as constants but we won't have to export/import them, they;ll be internal to this file

WDYT?

gregjopa:
> one alternative to constants is using derived properties.

@dreyks I like that idea! Let's go with that.

I saw this derived properties approach mentioned in the [Stop Using isLoading Booleans](https://kentcdodds.com/blog/stop-using-isloading-booleans) article you previously shared:

> if you really want your users to not have to use the variable === 'string', then just make sure you stick with the status in your state to ensure you only have one possible finite state value and then you can derive boolean states as a convenience if you want to

We should be all good with using derived properties since only `loadingStatus` is used to derive `isPending`, `isResolved`, and `isRejected` and there will always be only one possible value.

dreyks:
oh, so that's where I've got it from :D read the article a long time ago and looks like the ideas stuck in my head. kudos, Kent C. Dodds

I'll do the change

dreyks:
done. really like how it looks now

elizabethmv (approved):
Pulled down and confirmed locally. Looks good! 🏅

PR #52, opened 2020-11-01
Opt-out from using onShippingChange() callback with storybook demo

### LINKED ISSUE #50: [Bug] Card Fields instead of Popup Checkout flow
Hi guys, didn't know if this qualified as a bug or not, but I'm getting some weird behaviour from the "Debit or Credit Card" button, or at least inconsistent with the examples shown in the Storybook.

I've implemented everything as suggested in the docs, using all the defaults and whenever I click on the "Debit or Credit Card" button I'm getting the following form:

Captura de Pantalla 2020-10-29 a la(s) 01 50 23

which animates in, instead of it taking me to the popup checkout flow as it does in the storybook example. (Shown below)

Captura de Pantalla 2020-10-29 a la(s) 01 52 19

I'm wondering if this is the intended behaviour and if so, how could I revert it to the popup flow shown in the storybook examples? I couldn't find any info about this on the documentation.

FYI, I'm running v1.0.3

Sorry if this is not the appropriate forum for this question and thanks in advance for any help.

gregjopa:
This PR fixes #50 and makes sure the default inline guest feature works with the Standard Cards button for the [Buttons Storybook Demo](https://paypal.github.io/react-paypal-js/?path=/docs/example-paypalbuttons--default). Ex:

Screen Shot 2020-11-01 at 5 29 13 PM

This bug was only for the storybook demos. It turns out the storybook docs addon passes default values for props for component stories like the Buttons story even though we are not passing default prop type values. Passing default functions to most callbacks doesn't matter. But with the `onShippingChange` callback it does matter and using that option forces the user into the popup experience with the Standard Cards button. This PR passes null instead of a noop function to `onShippingChange` so the inline guest functionality works as expected.

### DIFF
--- a/src/stories/PayPalButtons.stories.js
+++ b/src/stories/PayPalButtons.stories.js
@@ -8,6 +8,12 @@ export default {
         shippingPreference: { control: null },
         style: { control: null },
     },
+    args: {
+        // Storybook passes empty functions by default for props like `onShippingChange`.
+        // This turns on the `onShippingChange()` feature which uses the popup experience with the Standard Card button.
+        // We pass null to opt-out so the inline guest feature works as expected with the Standard Card button.
+        onShippingChange: null,
+    },
 };
 
 const defaultOptions = {
--- a/src/stories/PayPalMarks.stories.js
+++ b/src/stories/PayPalMarks.stories.js
@@ -73,7 +73,10 @@ function RadioButtonTemplate(args) {
                 ))}
             </form>
             <br />
-            <PayPalButtons fundingSource={fundingSource} />
+            <PayPalButtons
+                fundingSource={fundingSource}
+                style={{ color: "white" }}
+            />
         </PayPalScriptProvider>
     );
 }
--- a/src/stories/Subscriptions.stories.js
+++ b/src/stories/Subscriptions.stories.js
@@ -59,12 +59,12 @@ function TransactionTypeForm() {
                 value: scriptOptions,
             });
         } else {
+            const { "client-id": clientId, components } = scriptOptions;
             dispatch({
                 type: "resetOptions",
                 value: {
-                    ...scriptOptions,
-                    intent: "order",
-                    vault: false,
+                    "client-id": clientId,
+                    components,
                 },
             });
         }


src/stories/PayPalButtons.stories.js
```diff
@@ -8,6 +8,12 @@ export default {
         shippingPreference: { control: null },
         style: { control: null },
     },
+    args: {
+        // Storybook passes empty functions by default for props like `onShippingChange`.
+        // This turns on the `onShippingChange()` feature which uses the popup experience with the Standard Card button.
+        // We pass null to opt-out so the inline guest feature works as expected with the Standard Card button.
+        onShippingChange: null,
```
elizabethmv, line 15: Can we add this change to the Subscriptions demo and the Marks radio button demo which also use the standard card button?

src/stories/PayPalButtons.stories.js
```diff
@@ -16,6 +22,7 @@ const defaultOptions = {
 };
 
 function Template(args) {
+    console.log("args", args);
```
elizabethmv, line 25:
```suggestion
```

elizabethmv (commented):
When testing this locally I'm getting a mix of both. Sometimes I get the popup sometimes I get inline 🤔 
| Popup        | Inline           |
| -| -|
|Screen Shot 2020-11-02 at 9 28 32 AM|Screen Shot 2020-11-02 at 9 28 51 AM|

src/stories/PayPalButtons.stories.js
gregjopa: Thanks for reviewing and testing this locally! I think this `onShippingChange` change is needed only for the buttons story because we export the component like so:

```js
export default {
    title: "Example/PayPalButtons",
    component: PayPalButtons,
```

And then use the component args like so:
```jsx
function Template(args) {
    return (
        <PayPalScriptProvider options={defaultOptions}>
            <PayPalButtons {...args} />
        </PayPalScriptProvider>
    );
}
```

Those args passed to the `<PayPalButtons />` contain all the inferred default values because the component is exported. The `<PayPalMarks />` component only has `fundingSource` as the param so we don't need to worry about it. And the Subscriptions story doesn't actually export a component so it doesn't need it.

gregjopa:
@elizabethmv I pushed up a couple more commits. Please test this again locally. 

> When testing this locally I'm getting a mix of both. Sometimes I get the popup sometimes I get inline 🤔

When testing locally here's what to look for:

1. All the example paypal button stories should open the inline guest form. I noticed this was not the case from your screenshot. Please clear browser cache and try again. I wasn't able to reproduce on my end. If it still happens for you please let me know and I will dig into it more.
2. The subscriptions example always uses the popup approach. The subscription functionality really depends on the buyer signing in and doesn't work with the buyer being a guest. The toggle demo with subscriptions should use the inline guest form when Order is selected. There was a bug related to "intent=order" no longer being a valild query param. I fixed that and it's working again:
Screen Shot 2020-11-02 at 10 25 48 AM
3. The Marks radio button example was broken. It seems like the default color for the Marks standard card button is gold which is breaking it. Explicitly setting all the button colors to white fixes the issue.
Screen Shot 2020-11-02 at 10 25 26 AM

elizabethmv (approved):
> All the example paypal button stories should open the inline guest form. I noticed this was not the case from your screenshot. Please clear browser cache and try again. I wasn't able to reproduce on my end. If it still happens for you please let me know and I will dig into it more.

@gregjopa I pulled the latest changes and tested locally again. After looking a second time, it seems that this issue only happens when switching between stories (even with cleared cache). I was able to reproduce it again by switching between Default, Dynamic, or Custom Style stories and clicking the card button.

Screen Shot 2020-11-02 at 11 17 01 AM

That said, I don't think this is a blocker for merge, so I'll leave that up to you. Thanks for addressing my questions and for the clear description!

gregjopa:
@elizabethmv I see what you mean now. I can reproduce the inconsistent popup behavior. I tested in Chrome and it seems to open up the popup the most compared to other browsers. It's definitely inconsistent. I think this PR is still a good improvement so I'm going to merge it.

PR #56, opened 2020-11-06
Add forceReRender prop to the Buttons to control the re-render of the…

yocmen:
This is a little hacky but it work, the idea is to control the re-render of the component passing our changing state to the component to force the re-render and take the new value/s if we need it, and it will not re-render the PayPalButtons if we pass methods wich re-render our parent component.

I added the example to the dynamic Amount in the stories.

Hope this works or gives you some idea on how we can make it work. 👍 

Thanks.


gregjopa:
@yocmen awesome work! I like the simpleness of this design. By default, most use cases don't need to trigger a re-render of the buttons so we should be avoid tracking all prop changes in the `useEffect()` hook. We want folks to be able to manage there own state and update it using the SDK callbacks (createOrder, onApprove, etc) without trigger a rerender of the buttons and this change enables that 💯

src/components/PayPalButtons.js
```diff
@@ -141,6 +141,10 @@ PayPalButtons.propTypes = {
      * Called when the buyer changes their shipping address on PayPal.
      */
     onShippingChange: PropTypes.func,
+    /**
+     * Called when the button need to be re-rendered.
```
gregjopa, line 145:
```suggestion
     * Used to re-render the component. Changes to this prop will destroy the existing Buttons and render them again using the current props.
```

src/components/PayPalButtons.js
```diff
@@ -141,6 +141,10 @@ PayPalButtons.propTypes = {
...
+     */
+    forceReRender: PropTypes.string,
```
gregjopa, line 147: I don't think we need to limit this to a string. I think a number or boolean or any other value would be fine too.
```suggestion
    forceReRender: PropTypes.any,
```

src/stories/PayPalButtons.stories.js
```diff
@@ -93,17 +93,23 @@ StandAlone.parameters = {
 
 function DynamicAmountTemplate(args) {
     const [amount, setAmount] = useState(2);
+    const [orderID, setOrderID] = useState();
```
gregjopa, line 96: I think it would be nice to have a default value for Order ID and reset it when the amount changes.
```suggestion
    const [orderID, setOrderID] = useState('unknown');
```

src/stories/PayPalButtons.stories.js
```diff
@@ -93,17 +93,23 @@ StandAlone.parameters = {
...
+                        amount: {
+                            value: amount,
+                        },
                     },
-                },
-            ],
-        });
+                ],
+            })
+            .then((orderId) => {
```
gregjopa, line 109: Could we update this to `orderID`? That way all the code is consistent with using `ID` instead of `Id`

src/stories/PayPalButtons.stories.js
```diff
@@ -113,18 +119,24 @@ function DynamicAmountTemplate(args) {
...
-                <option value="6">$6.00</option>
-            </select>
-
-            <PayPalButtons {...args} createOrder={createOrder} />
+            <div>
+                <select
+                    onChange={onChange}
+                    name="amount"
+                    id="amount"
+                    style={{ marginBottom: "20px" }}
```
gregjopa, line 127:
```suggestion
```

src/stories/PayPalButtons.stories.js
```diff
@@ -113,18 +119,24 @@ function DynamicAmountTemplate(args) {
...
+                >
+                    <option value="2">$2.00</option>
+                    <option value="4">$4.00</option>
+                    <option value="6">$6.00</option>
+                </select>
+                &nbsp;Order ID: {orderID}
```
gregjopa, line 133: Let's using a paragraph tag which will help with the vertical spacing 
```suggestion
                <p>Order ID: {orderID}</p>
```

src/stories/PayPalButtons.stories.js
```diff
@@ -113,18 +119,24 @@ function DynamicAmountTemplate(args) {
...
-            >
-                <option value="2">$2.00</option>
-                <option value="4">$4.00</option>
-                <option value="6">$6.00</option>
-            </select>
-
-            <PayPalButtons {...args} createOrder={createOrder} />
+            <div>
+                <select
+                    onChange={onChange}
```
gregjopa, line 124: Let's call `setOrderID('unknown')` in the onChange function so the UI reflects when the order is reset.

gregjopa:
@yocmen I left some minor comments with requests for changes. Please let us know if you'd be up for making them. 

It would be great to add a couple unit tests for this behavior:
1. Test that `forceReRender` calls the `window.paypal.Buttons().render()` mock again when triggering a re-render
2. Test that other props do not trigger a re-render. I'm not exactly sure what this unit test would look like. Perhaps we just choose a couple prop callbacks like` createOrder` and `onApprove` and ensure changes to them do not trigger a re-render.

Also, please do not check in the yarn.lock file since we aren't using yarn with this project.

src/components/PayPalButtons.js
```diff
@@ -45,7 +45,7 @@ export default function PayPalButtons(props) {
         });
 
         return cleanup;
-    });
+    }, [isResolved, props.forceReRender]);
```
gregjopa, line 48: We do have a couple other prop values that are primitives that should probably trigger a re-render. 
```suggestion
    }, [isResolved, props.forceReRender, props.fundingSource, props.shippingPreference]);
```

yocmen:
Perfect @gregjopa I will work on all that.

src/stories/PayPalButtons.stories.js
mnicpt: @gregjopa Why not set the default value to something falsy rather than `unknown`?

gregjopa: @mnicpt good call. Setting it to false seems like a better design. Then we can conditionally show the "unknown" label in jsx like so:
```jsx
<p>Order ID: {orderID ? orderID : 'unknown' }</p>
```

giladv:
just wanted to add that this fix fixes a critical bug were the component that renders the PayPalButtons cannot change nothing while being displayed. and i mean even an internal state change, because the effect has no dependencies array so it tries to re-render the buttons on every change on the parent component unless you memoize the function props. 
this need to be pushed as soon as possible. thanks:)

yocmen:
Hi, I made all the changes but I'm having problems with the tests, if anyone can help with that could be great :)

gregjopa:
Hi @yocmen, I’m happy to help with the tests. Can you push up your most recent changes?

yocmen:
done, please check :)

.gitignore
```diff
@@ -3,3 +3,4 @@ dist
 
 node_modules
 .DS_Store
+yarn.lock
```
gregjopa, line 6: nit: Please keep the new line at the end of the file

src/components/PayPalButtons.js
```diff
@@ -45,7 +45,12 @@ export default function PayPalButtons(props) {
         });
 
         return cleanup;
-    });
+    }, [
+        isResolved,
+        props.forceReRender,
+        props.fundingSource,
+        props.shippingPreference,
```
gregjopa, line 52: Thanks for updating this. Sorry to bug you with another change but could you also add the style prop like so?

```suggestion
        props.shippingPreference,
        props.style,
```

gregjopa:
@yocmen the new changes look great! I left a couple minor comments.

> I'm having problems with the tests, if anyone can help with that could be great :)

For the unit tests, I added a failing test for this issue in PR #57. I do not plan on merging that PR. Instead, please copy that new test over to your code and commit it and push it up. It should pass 🤞. 

Can you try adding an additional test for the `forceReRender` prop to confirm it does trigger a re-render? You should be able to use a very similar approach. After that, this PR should be ready for a final review before merging.

gregjopa:
@yocmen I'm going to merge this since I'd like to cut a 3.0 release later today. I'll take care of the unit tests. Thank you so much for your contributions! 😄 💯 🥇

yocmen:
Ok, sorry for the delay, I was looking at the test and "playing" with the new tests, but glad to help a little :) really thanks.

PR #61, opened 2020-11-17
Memoize PayPalButtons component

gregjopa:
Experiment with the idea of using `memo` to prevent side-effects from triggering a re-render and breaking the iframe.

- https://stackoverflow.com/a/52874606
- https://reactjs.org/docs/react-api.html#reactmemo


pedroapfilho:
This is not necessary, but it can fix a problem that we don't have yet, and it doesn't make it worse. I think this is a good catch

gregjopa:
@pedroapfilho I think being very careful with re-renders is a good thing and we will continue to do that with this library. I'm going to close out this draft PR for now since it isn't solving any particular bug. We can re-visit this in the future if more re-render issues arise.

PR #67, opened 2020-11-25
Add a tiny button example

gregjopa:
This PR adds a new story for rendering a standalone button as small as possible. It also simplifies the storybook implementation for Buttons.

Screen Shot 2020-11-25 at 1 42 12 PM

I also updated the "Stand alone" button story to use PayPal instead of PayLater since the PayLater button may not show to some users based on their country.

### DIFF
--- a/src/stories/PayPalButtons.stories.js
+++ b/src/stories/PayPalButtons.stories.js
@@ -80,12 +80,38 @@ CustomStyle.parameters = {
 };
 
 export const StandAlone = Template.bind({});
-StandAlone.args = { fundingSource: FUNDING.PAYLATER };
+StandAlone.args = { fundingSource: FUNDING.PAYPAL };
 
 StandAlone.parameters = {
     docs: {
         source: {
-            code: "<PayPalButtons fundingSource={FUNDING.PAYLATER} />",
+            code: "<PayPalButtons fundingSource={FUNDING.PAYPAL} />",
+        },
+    },
+};
+
+function TinyTemplate(args) {
+    return (
+        <PayPalScriptProvider options={defaultOptions}>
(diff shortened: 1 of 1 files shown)


src/stories/PayPalButtons.stories.js
```diff
@@ -80,12 +80,38 @@ CustomStyle.parameters = {
...
-            code: "<PayPalButtons fundingSource={FUNDING.PAYLATER} />",
+            code: "<PayPalButtons fundingSource={FUNDING.PAYPAL} />",
+        },
+    },
+};
+
+function TinyTemplate(args) {
+    return (
+        <PayPalScriptProvider options={defaultOptions}>
+            <div style={{ width: "75px" }}>
```
gregjopa, line 96: This wrapping div is forcing the container to have a width of 75px which seems to be the smallest the paypal button can go.

src/stories/PayPalButtons.stories.js
```diff
@@ -13,110 +18,38 @@ export default {
         // We pass null to opt-out so the inline guest feature works as expected with the Standard Card button.
         onShippingChange: null,
     },
+    decorators: [
```
gregjopa, line 21: I also simplied the storybook implementation for Buttons in this PR. I learned that we can use the [storybook decorator](https://storybook.js.org/docs/react/writing-stories/decorators) to wrap every story in the `<PayPalScriptProvider>`. The nice thing about this is the provider is not included in the code preview snippets. And by removing the complexity with `bind` and `args`, we now can just use the actual story code in the preview 😄 

Screen Shot 2020-11-27 at 2 51 07 PM



Screen Shot 2020-11-27 at 2 51 07 PM

Issue #69, opened 2020-12-04
[Feature] Use TypeScript


gregjopa:
### 🚀 Feature Proposal

Update this library to use [TypeScript](https://www.typescriptlang.org/).

### Motivation

Several developers have requested TypeScript types for this library. We could solve that problem in a few different ways. One way is to convert this codebase to TypeScript and include types for all publicly exported functions.

Here's a good doc about [the advantages of TypeScript](https://www.typescriptlang.org/why-create-typescript).

### Related PRs

#30 - @camsjams shared some really helpful types here which would be a good starting point: https://gist.github.com/camsjams/b843637df7a30c260b341daf97a96d60
#53

### Additional Info

The current use of PropTypes is powering the Storybook docs: https://paypal.github.io/react-paypal-js/?path=/docs/example-paypalbuttons--default. We want to continue supporting PropTypes like this with using TypeScript. That may be one of the more challenging parts of the migration.

Any objections to migrating to TypeScript?

Any volunteers to take this work up?

sahaparamjit:
@gregjopa I would like to work on this. I am learning TypeScript. So I guess it would be great if I take that.

gregjopa:
@sahaparamjit that would be awesome! Happy to answer any questions you may have along the way.

gregjopa:
Hi @sahaparamjit just wanted to check in. Have you had a chance to work on this? Let me know if you have any questions.

isaacewing:
@gregjopa  I would also like to work on this. you made a great component... in javascript ha. I'm pretty good with typescript and would like to help. right now, i'm creating interfaces for my project and would like to contribute them the ts version of this when it's ready

gregjopa:
@isaacewing that sounds great! I'd appreciate the help.

HunteRoi:
Hey there! I know this started a few days ago but would it be possible to have some updates? I am working on a React+Typescript project and I would be really happy to see this library updated to its TS version :smile:

gregjopa:
Hello TypeScript friends! I started hacking on TypeScript for paypal-js and would appreciate a code review if anyone has time: https://github.com/paypal/paypal-js/pull/55. The idea is to add TypeScript to paypal-js and then to this library (which depends on paypal-js).

The idea with this design is paypal-js would provide type definitions for the whole JS SDK so everyone will know what's possible with `window.paypal`. I also think the rollup config work from this PR could be applied to react-paypal-js.

HunteRoi:
[...]

Issue #53, opened 2020-11-02
TypeScript Issue - usually optional script options are required


giladv:
usually only the client-id is required. but it seems all of the options are required in this package
```
TS2740: Type '{ 'client-id': string; intent: string; }' is missing the following properties from type
 'Required<InferProps<{ "buyer-country": Requireable<string>; "client-id": Validator<string>; 
commit: Requireable<string | boolean>; components: Requireable<string>; ... 12 more ...; 
vault: Requireable<...>; }>>': "buyer-country", commit, components, currency, and 11 more.
ℹ ｢wdm｣: Failed to compile.
```

how to reproduce:
1. basic typescript setup
2.
 ```
<PayPalScriptProvider options={{ 'client-id': 'some id', intent: 'authorize' }}></PayPalScriptProvider>
```

**Comment by gregjopa** [comment]
Hi @giladv we have an open issue for adding types for this library: https://github.com/paypal/react-paypal-js/issues/30. Once those types are added, it should make it easy to use react-paypal-js with typescript. 

What typescript setup are you using? I tried to reproduce using a new create-react-app project using the typescript template. I get the following compile error about missing types:

```
TypeScript error in /Users/gjopa/projects/my-typescript-app/src/App.tsx(5,53):
Could not find a declaration file for module '@paypal/react-paypal-js'. '/Users/gjopa/projects/my-typescript-app/node_modules/@paypal/react-paypal-js/dist/react-paypal.node.js' implicitly has an 'any' type.
  Try `npm install @types/paypal__react-paypal-js` if it exists or add a new declaration (.d.ts) file containing `declare module '@paypal/react-paypal-js';`  TS7016

    3 | import './App.css';
    4 | import './index.d.ts';
  > 5 | import { PayPalScriptProvider, PayPalButtons } from "@paypal/react-paypal-js";
      |                                                     ^
    6 |
    7 | function App() {
    8 |   return (
```

**Comment by giladv** [comment]
hi @gregjopa, here is my tsconfig file
```
{
	"compilerOptions": {
		"target": "es5",
		"lib": ["dom", "dom.iterable", "esnext"],
		"allowJs": true,
		"noImplicitAny": false,
		"skipLibCheck": true,
		"esModuleInterop": true,
		"experimentalDecorators": true,
		"allowSyntheticDefaultImports": true,
		"strict": false,
		"forceConsistentCasingInFileNames": true,
		"module": "esnext",
		"moduleResolution": "node",
[...]

PR #71, opened 2021-01-10
Use TypeScript

gregjopa:
Happy new year everyone! I come bearing the gift of types 🎁 

### Screenshots

script-options-screenshot

buttons-screenshot

### How to Test

Use the latest beta to test out the types:

```
npm install @paypal/react-paypal-js@beta
```

### Additional Info

There are a lot of `@ts-expect-error` statements being used to avoid compile time errors. These are code smells and should be cleaned up before merging this PR. I could use some help with this. I would gladly accept PRs to this "typescript" branch.

I decided to keep the Proptypes with the idea that they are for runtime validation and the TypeScript types are for compile time validation. For more info see: https://stackoverflow.com/a/54690878

Related issues: #69 #53

### DIFF
--- a/src/components/PayPalButtons.tsx
+++ b/src/components/PayPalButtons.tsx
@@ -1,6 +1,11 @@
 import React, { useEffect, useRef, useState } from "react";
 import PropTypes from "prop-types";
 import { usePayPalScriptReducer } from "../ScriptContext";
+import type { PayPalButtonsComponentProps } from "@paypal/paypal-js/types/components/buttons";
+
+interface PayPalButtonsReactProps extends PayPalButtonsComponentProps {
+    forceReRender?: unknown
+}
 /**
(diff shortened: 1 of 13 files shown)


src/components/PayPalButtons.tsx
```diff
@@ -20,6 +25,7 @@ export default function PayPalButtons(props) {
     useEffect(() => {
         const cleanup = () => {
             if (buttons.current) {
```
mnicpt, line 27: Should we check for `buttons` being initialized here as well? `if (buttons && buttons.current`

mnicpt (approved):
Looks good, @gregjopa !  LGTM upon cleanup of `@ts-expect-error - null checks`

gregjopa:
Update: I was able to remove all the `@ts-expect-error` comments from the Buttons component. Hoping to do the same to the rest of the components and merge this PR next week.

Issue #85, opened 2021-02-24
[Bug] Failed to render <PayPalButtons /> component. Error: zoid destroyed all components


dan003400:
We are randomly getting this error on page refresh, it's about 50/50 if it works on each page load.

The error message makes no sense nor is it helpful, so we have no idea where to even look to diagnose the issue.

yocmen:
Can you provide minimal steps to reproduce your component? maybe some codesandbox 😁

dan003400:
This is happening a minimal setup using the context provider and the buttons component.

What does this error mean? What causes this error to happen?

Being that it's random, is this a network timing issue?

This issue does not happen with third party packages that manage the loading of your script and rendering of the button, only the official package from PayPal has issues.

We made the switch to avoid errors from strict mode.

dan003400:
Here is another guess.

The script starts to load, something in parent components causes a re-render while the PayPal buttons have started to load but have not completed.

Sounds like people have been getting this error message with PayPal for years, specially when there is two buttons, this makes me believe this has to do with scripts being loaded asynchronously while rendering is also being done asynchronously and there is a timing mismatch of some kind?

Regardless this error message is terrible.

dan003400:
Is this related this this?

https://github.com/krakenjs/zoid/pull/335

gregjopa:
Hi @dan003400, this `zoid destroyed all components` error happens when re-rendering the buttons without closing the existing buttons on the page. Zoid is the iframe component library we use and this error comes from it. It's usually a sign that the implementation code is unexpectedly causing a re-render. Please share an example

dan003400:
Im not sure how we can share an example. This is mounted in a large app, what would you need to diagnose this issue.

We have recently memoized the PayPalButtons component and have confirmed it will only render once with logs, and this issue will still happen 3/10 page loads.

dan003400:
For some context, this package is working flawlessly with the exact same implementation in the same location.

https://github.com/Luehang/react-paypal-button-v2

dan003400:
I'm not sure I understand why this has been closed?

catchacourse:
> For some context, this package is working flawlessly with the exact same implementation in the same location.
> 
> https://github.com/Luehang/react-paypal-button-v2

I'am using
https://github.com/Luehang/react-paypal-button-v2
and encountering exactly the same issue when using more than one dynamically created paypal button.

dan003400:
Has this been fixed or ignored via the early close?

PR #88, opened 2021-03-11
Do not throw button render errors when ref is invalid

gregjopa:
This PR attempts to limit rendering errors related to the Buttons container being removed before async rendering completes. It's inspired by this PR for the zoid react driver that ships with the JS SDK: https://github.com/krakenjs/zoid/pull/335.

Related: #85


gregjopa:
This empty container check was also needed: https://github.com/paypal/react-paypal-js/commit/b85932bd1efda453c3b3b952554558d288b67857

Issue #49, opened 2020-10-28
[Feature] Hosted Fields component


mglaman:
### 🚀 Feature Proposal

It'd be great to have a hosted fields component and support: https://developer.paypal.com/docs/business/checkout/advanced-card-payments/#

Currently, you cannot set the client-token as an option on the PayPalProvider. So even trying to write a custom component will not work.

### Motivation

We want to embed hosted fields

### Example

We've implemented this for the Drupal Commerce PayPal integration and are adding a headless example.

mglaman:
Oh, just because it isn't in the props types didn't mean it was supported.

```
      <PayPalScriptProvider options={{
        'components': 'hosted-fields,buttons',
        'client-id': process.env.REACT_APP_PAYPAL_SB_CLIENT_ID,
        'data-partner-attribution-id': 'CommerceGuys_Cart_SPB',
        'data-client-token': process.env.REACT_APP_PAYPAL_SB_CLIENT_TOKEN,
      }}>
```

With this we can detect if eligible and use a hosted field setup or fallback to smart buttons

```
  const [{ isLoaded, options }] = usePayPalScriptReducer();
  const [supportsHostedFields, setSupportsHostedFields] = useState(null);

  useEffect(() => {
    console.log(options);
    if (isLoaded) {
      setSupportsHostedFields(window.paypal.HostedFields.isEligible())
    }
  }, [isLoaded])
```

gregjopa:
@mglaman thanks for opening up this issue. You are correct that `data-client-token` should work fine even though it's missing from the prop-types. Thanks for pointing that out. I added it to the prop-types in PR #51.

Hosted-fields should definitely be possible using the custom component approach you outlined in https://github.com/paypal/react-paypal-js/issues/49#issuecomment-718198987. 

Another thing to note is the upcoming 2.0 release drops the `isLoaded` option and replaces it with `isPending`, `isResolved`, and `isRejected`. Ex:

```js
const [{ isResolved, options }] = usePayPalScriptReducer();
  const [supportsHostedFields, setSupportsHostedFields] = useState(null);

  useEffect(() => {
    console.log(options);
    if (isResolved) {
      setSupportsHostedFields(window.paypal.HostedFields.isEligible())
    }
  }, [isResolved])
```

I'm going to close this one out. Please re-open if you encounter any problems with the hosted-fields implementation.

getTobiasNielsen:
Hey @gregjopa

Thank you for your great work for react community.

[...]

PR #95, opened 2021-03-24
Initial prototype

getTobiasNielsen:
A lot of the types are missing, would need to implement those, either in the file or in their respective places.

https://github.com/paypal/react-paypal-js/issues/49

This is an initial prototype.

I was thinking it could be smart to add one of the default spinners I've seen paypal use, or if people don't like it, let people add a prop with a spinner element or some custom react component, if they fancy that instead.

### DIFF
--- a/src/components/PayPalHostedFields.tsx
+++ b/src/components/PayPalHostedFields.tsx
@@ -1,0 +1,136 @@
+import React, {
+    useEffect,
+    useRef,
+    useState,
+    ReactElement,
+} from "react";
+import { usePayPalScriptReducer } from "../ScriptContext";
+import { getPayPalWindowNamespace, DEFAULT_PAYPAL_NAMESPACE } from "./utils";
+import type {
+    PayPalHostedFieldsComponentProps,
+    PayPalHostedFieldsComponent,
+    OnInitActions,
+} from "@paypal/paypal-js/types/components/HostedFields";
+
+export interface PayPalHostedFieldsProps extends PayPalHostedFieldsComponentProps {
+    /**
+     * Used to re-render the component.
+     * Changes to this prop will destroy the existing HostedFields and render them again using the current props.
+     */
+     forceReRender?: unknown;
+     /**
+    /**
+     * Pass a css class to the div container.
+     */
+    className?: string;
+    /**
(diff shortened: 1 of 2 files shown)


src/components/PayPalHostedFields.tsx
```diff
@@ -0,0 +1,136 @@
...
+        // billing address here
+ *     </PayPalHostedFields>
+ * ```
+ */
+
+export const PayPalHostedFields: React.FunctionComponent<PayPalHostedFieldsProps> = ({className = "", forceReRender, children, ...hostedFieldsProps}) => {
+    // const hostedFields = useRef<PayPalHostedFieldsComponent | null>(null);
+    const hostedFields = useRef<any | null>(null);
+    const [{ isResolved, options }] = usePayPalScriptReducer();
+    const [initActions, setInitActions] = useState<OnInitActions | null>(null);
```
gregjopa, line 51: This `initActions` code is unique to the Buttons component. We shouldn't need it for HostedFields.

src/components/PayPalHostedFields.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+// useEffect hook for rendering the hosted-fields
+useEffect(() => {
+    // verify the sdk script has successfully loaded
+    if (isResolved === false) {
+        return closeHostedFieldsComponent;
+    }
+
+    // Make sure a client token is provided
+    if (options["data-client-token"] === null) {
```
gregjopa, line 71: I'm wondering how this "data-client-token" value would get set in someone's integration code. Can you share an example of what that would look like with the PayPalScriptProvider and this new PayPalHostedFields component? Maybe as a storybook example?

gregjopa:
Thanks for the PR @getTobiasNielsen! I think this is a great start! 💯 I added a couple comments.

src/components/PayPalHostedFields.tsx
getTobiasNielsen: Yeah, this is interesting. Right now, I am using the dispatch to set it, but I imagine going forward, we can make a client id loader, that will take care of fetching an api that is ready to deliver the client-token.

I think having this separate from hosted-fields, will be more helpful. An alternative, is to give the paypal script provider, a route to fetch.

PR #97, opened 2021-03-29
Adding example for useEffect changing the currency

### LINKED ISSUE #96: [Bug] Currency change not reliable
### 🐞 Describe the Bug

It seems that if you add the `dispatch` on the `useEffect` to change the currency, and it triggers 2 times fastly, it doesn't get the correct currency reliably. 

### 🔬 Minimal Reproduction

We have a context, and we have a `defaultCurrency`, and a `selectedCurrency`, that is loaded from the server, so it initializes as `GBP` (default currency), and then fastly goes to `EUR` (as an example of selected currency that came from the server, on an API call)

### 😕 Actual Behavior

Creates the order as `GBP`

### 🤔 Expected Behavior

Should create the order in `EUR`

### ➕ Additional Context

What we did to fix this race condition was adding a check on the `useEffect`, like this:

```js
  useEffect(() => {
    if (selectedCurrency !== defaultCurrency) {
      paypalDispatch({
        type: "resetOptions",
        value: {
          ...options,
          currency: selectedCurrency,
        },
      });
    }
  }, [selectedCurrency]);
```

Then it works reliably.

So either we have to add it to the documentation, or create a solution for it

pedroapfilho:
Fixes #96 

https://user-images.githubusercontent.com/13142568/112827255-75a7e980-9064-11eb-92bc-84d9e8d9c744.mov

### DIFF
--- a/src/stories/usePayPalScriptReducer.stories.js
+++ b/src/stories/usePayPalScriptReducer.stories.js
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import {
     PayPalScriptProvider,
     usePayPalScriptReducer,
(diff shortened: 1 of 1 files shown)


gregjopa:
I’m not a fan of this useEffect solution since it’s ignoring the state of the asynchronous script loader. How about we leverage the disabled state of the select box instead? The thinking is the select box remains disabled while the script is loading based on isResolved state.

pedroapfilho:
Well, I completely understand that this example doesn't quite shows the benefits of it, but this example was created to show how to do it for devs who actually need it. In my case, was when I had a `CurrencyContext`, and changed the the `currency` based on the user location, so we load first it as `GBP`, and then to whatever the user needs.

What do you suggest? Creating a full usable example?

gregjopa:
Closing this one out in favor of https://github.com/paypal/react-paypal-js/pull/100 instead.

PR #100, opened 2021-03-31
fix: prevent script provider from loading more than one sdk script

### LINKED ISSUE #96: [Bug] Currency change not reliable

gregjopa:
This PR updates the behavior around the "resetOptions" action to clean up any old scripts before reloading. This should fix #96.

As @pedroapfilho pointed out, multiple scripts can get triggered to load if you dispatch multiple "resetOptions" actions before the previous script finishes loading. This is easier to reproduce when throttling your network connection. Here's a screenshot of the problem:

multiple-scripts-loading-issue

This PR fixes the issue by introducing a data attribute to uniquely identify the script element. And to clean it up whenever "resetOptions" action is triggered before loading another script.

This PR has been published to npm as `@paypal/react-paypal-js@5.1.1-beta.1`.

@pedroapfilho can you test this beta version out and let me know if it fixes the issue? With this change you shouldn't need that useEffect loop comparing old and new currencies anymore.

### DIFF
--- a/src/ScriptContext.test.js
+++ b/src/ScriptContext.test.js
@@ -7,9 +7,26 @@ jest.mock("@paypal/paypal-js", () => ({
     loadScript: jest.fn(),
 }));
 
+function loadScriptMockImplementation({
+    "client-id": clientID,
+    "data-react-paypal-script-id": reactPayPalScriptID,
+}) {
+    const newScript = document.createElement("script");
+    newScript.src = `https://www.paypal.com/sdk/js?client-id=${clientID}`;
+    newScript.setAttribute("data-react-paypal-script-id", reactPayPalScriptID);
+
+    document.head.insertBefore(newScript, document.head.firstElementChild);
+    return Promise.resolve({});
+}
+
 describe("<PayPalScriptProvider />", () => {
     beforeEach(() => {
-        loadScript.mockResolvedValue({});
+        document.head.innerHTML = "";
+        loadScript.mockImplementation(loadScriptMockImplementation);
+    });
+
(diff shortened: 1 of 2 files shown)


pedroapfilho:
I'll test it out tomorrow morning and ping you out! I'm in debt with my other PR, working on it!

pedroapfilho:
It works! Basically, it only loads once and makes it reliably work

gregjopa:
> It works! Basically, it only loads once and makes it reliably work

Thanks for testing this out @pedroapfilho! I'll move forward with merging and publishing a release.

PR #101, opened 2021-04-02
feat: add deferLoading option to control when sdk script loads

### LINKED ISSUE #89: [Feature] Add option to control when the JS SDK script loads
### 🚀 Feature Proposal

Add the ability to control when the `<PayPalScriptProvider>` loads the JS SDK `<script>`. Currently it loads the JS SDK script as soon as the provider is rendered. 

### Motivation

In complex react apps, you may only want to load the JS SDK script under certain conditions.

### Example

My initial thought is to add a new prop to the `<PayPalScriptProvider>` to control this behavior:

```jsx
<PayPalScriptProvider options={{ "client-id": "test" }} deferLoading={true} >
    { /* pretend this component does not get shown on the initial route, and instead gets added later in the user flow */ }
    <PayPalButtons style={{ layout: "horizontal" }} />
</PayPalScriptProvider>
```
 
The ~`shouldLoadScript`~ `deferLoading` prop would accept a boolean value and default to false. Let's consider this behavior "eager loading". 

When true is passed in, the script loading behavior would be considered "lazy loading". In this mode you could load the script in a few different ways:
1. Update the value passed into the `deferLoading` prop and set it equal to false.
2. Wait for the the component (ex: `<PayPalButtons />`) to get rendered which would dispatch an action to kick off the script loading by changing the state to pending.
3. Dispatch the `setLoadingStatus` action to change from `NOT_STARTED` to `PENDING` to kick off the script loading.

Currently we have an [enum for the different script loading states](https://github.com/paypal/react-paypal-js/blob/main/src/ScriptContext.tsx#L5-L9). I think we would need to add a new "NOT_STARTED" option to support the `shouldLoadScript` prop:

```typescript
enum SCRIPT_LOADING_STATE {
    NOT_STARTED = "not_started",
    PENDING = "pending",
    REJECTED = "rejected",
    RESOLVED = "resolved",
}
```

gregjopa:
This PR adds a new `deferLoading` prop to the `<PayPalScriptProvider>` for controlling when the sdk script loads. Resolves #89. 

`deferLoading` prop behavior:
- set to false by default since we usually know all the sdk script params up front and want to load asap in order for the Buttons to render as fast as possible for buyers. 
- can be set to true to prevent loading the JS SDK script when the provider renders. Use `deferLoading={true}` initially and then dispatch an action later on in the app's life cycle to load the sdk script.

One thing to note is the `<PayPalScriptProvider>` props are currently designed to only support initial load. This new "deferLoading" prop supports prop changes. If you initially set it to true and then change it to false when rerendering, it will trigger a script load. The options prop still only works on initial load and the dispatcher is still the only way to make changes to it.

### DIFF
--- a/src/ScriptContext.tsx
+++ b/src/ScriptContext.tsx
@@ -13,6 +13,7 @@ export interface ReactPayPalScriptOptions extends PayPalScriptOptions {
 }
 
 enum SCRIPT_LOADING_STATE {
+    INITIAL = "initial",
     PENDING = "pending",
     REJECTED = "rejected",
     RESOLVED = "resolved",
@@ -25,6 +26,7 @@ interface ScriptContextState {
 
 interface ScriptContextDerivedState {
     options: ReactPayPalScriptOptions;
+    isInitial: boolean;
     isPending: boolean;
     isRejected: boolean;
     isResolved: boolean;
@@ -97,6 +99,7 @@ function usePayPalScriptReducer(): [
 
     const derivedStatusContext = {
         ...restScriptContext,
+        isInitial: loadingStatus === SCRIPT_LOADING_STATE.INITIAL,
         isPending: loadingStatus === SCRIPT_LOADING_STATE.PENDING,
         isResolved: loadingStatus === SCRIPT_LOADING_STATE.RESOLVED,
         isRejected: loadingStatus === SCRIPT_LOADING_STATE.REJECTED,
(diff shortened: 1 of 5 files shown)


src/ScriptContext.tsx
```diff
@@ -108,18 +112,22 @@ function usePayPalScriptReducer(): [
 interface ScriptProviderProps {
     options: PayPalScriptOptions;
     children?: React.ReactNode;
+    deferLoading?: boolean;
 }
 
 const PayPalScriptProvider: FunctionComponent<ScriptProviderProps> = ({
     options,
     children,
+    deferLoading = false,
```
gabrielo91, line 121: Since the boolean value of `undefined`  is false, I think the initial value of `false` in that line is not needed

gregjopa: Great point! Thanks for the review @gabrielo91

gabrielo91 (approved):

Issue #98, opened 2021-03-29
[Feature] forceRerender as an array


pedroapfilho:
Right now we have only one property on the `forceRerender` prop, I had to add my `createOrder` function to it, which is an `useCallback` function, and it respects the array of properties of it to force the rerender.

I think we can make it more readable/reliable if we make the `forceRerender` be an array, so we can pass as many properties to it as we want, and it would be very easy to implement.

gregjopa:
Hey @pedroapfilho, sure I'm open to adding this feature with turning the `forceRerender` prop into an array. Right now we expect this prop to be a primitive value so it would be a breaking change. Would you be up for sending a PR for adding this feature?

pedroapfilho:
I would! This change is very simple! I'll do it today at night!

PR #102, opened 2021-04-05
Changing forceReRender to an array

pedroapfilho:
Fixes https://github.com/paypal/react-paypal-js/issues/98

This makes the `forceReRender` an array, so we can add multiple things to the `useEffect` inside the component.

This enables us to add things like `currency`, `amount`, `order` and etc to the `forceReRender`, so it is always correct at the moment that the user clicks the button to Checkout.

This is a breaking change.

### DIFF
--- a/src/components/PayPalButtons.test.js
+++ b/src/components/PayPalButtons.test.js
@@ -108,7 +108,7 @@ describe("<PayPalButtons />", () => {
                     <button onClick={() => setAmount(amount + 1)}>
                         Update Amount
                     </button>
-                    <PayPalButtons forceReRender={amount} />
+                    <PayPalButtons forceReRender={[amount]} />
                 </>
             );
         }
--- a/src/components/PayPalButtons.test.js
+++ b/src/components/PayPalButtons.test.js
@@ -100,15 +100,19 @@ describe("<PayPalButtons />", () => {
         });
     });
 
-    test("should re-render Buttons when props.forceReRender changes", async () => {
+    test("should re-render Buttons when any item from props.forceReRender changes", async () => {
         function ButtonWrapper({ initialAmount }) {
             const [amount, setAmount] = useState(initialAmount);
+            const [currency, setCurrency] = useState("USD");
             return (
                 <>
                     <button onClick={() => setAmount(amount + 1)}>
                         Update Amount
                     </button>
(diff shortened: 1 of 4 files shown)


src/components/PayPalButtons.test.js
```diff
@@ -108,7 +108,7 @@ describe("<PayPalButtons />", () => {
                     <button onClick={() => setAmount(amount + 1)}>
                         Update Amount
                     </button>
-                    <PayPalButtons forceReRender={amount} />
+                    <PayPalButtons forceReRender={[amount]} />
```
gregjopa, line 111: Thanks @pedroapfilho! This looks great 💯 

It would be nice to have this unit test make sure `forceRerender` works with watching multiple variables. Could you update this to watch another variable? Something like:
```jsx
<PayPalButtons forceReRender={[amount, currency]} />
```
And then test currency the same way amount is being tested.

pedroapfilho: For sure I can! It was lazy from my side to only update these tests, I'll definitely add more!

pedroapfilho: Added!

One thing that I couldn't do was to run the project locally. It seems that I have some problems with `husky`. Do you have any place where you documented what you did on the update? And how you run this project?

gregjopa: Thanks for adding the additional tests! 

> One thing that I couldn't do was to run the project locally.

What operating system are you using? I'm on a mac and the following steps should work to get husky working:
1. `rm -rf node_modules`
2. `npm install`
3. `npm start` or `npm run validate`

Let me know if that helps.

src/components/PayPalButtons.test.js
```diff
@@ -100,15 +100,19 @@ describe("<PayPalButtons />", () => {
         });
     });
 
-    test("should re-render Buttons when props.forceReRender changes", async () => {
+    test("should re-render Buttons when any item from props.forceReRender changes", async () => {
```
gregjopa, line 103: 💯

src/components/PayPalButtons.test.js
gregjopa: I went ahead and merged the PR. The CI failure was just due to the commit message format. I fixed it with the squash and merge option.

Thanks for contributing @pedroapfilho! 💯 🥇 😄

PR #113, opened 2021-05-14
fix: typescript type for `ScriptReducerAction`

leogedler:
- Created `ScriptLoadingState` type from the union of the enum `SCRIPT_LOADING_STATE` and used it 
as a possible values for `setLoadingStatus` to avoid type error when passing a valid string value on the dispatch function.

### DIFF
--- a/src/ScriptContext.tsx
+++ b/src/ScriptContext.tsx
@@ -19,9 +19,11 @@ enum SCRIPT_LOADING_STATE {
     RESOLVED = "resolved",
 }
 
+type ScriptLoadingState = `${SCRIPT_LOADING_STATE}`;
+
 interface ScriptContextState {
     options: ReactPayPalScriptOptions;
-    loadingStatus: SCRIPT_LOADING_STATE;
+    loadingStatus: SCRIPT_LOADING_STATE | ScriptLoadingState;
 }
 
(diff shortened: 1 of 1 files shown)


src/ScriptContext.tsx
```diff
@@ -19,9 +19,11 @@ enum SCRIPT_LOADING_STATE {
     RESOLVED = "resolved",
 }
 
+type ScriptLoadingState = `${SCRIPT_LOADING_STATE}`;
```
gregjopa, line 22: @leogedler I was reading up on `enums` and I think we should get rid of it and replace it with the following:

```suggestion
const SCRIPT_LOADING_STATE = {
    INITIAL: "initial",
    PENDING: "pending",
    REJECTED: "rejected",
    RESOLVED: "resolved",
} as const;

type ScriptLoadingState = typeof SCRIPT_LOADING_STATE[keyof typeof SCRIPT_LOADING_STATE];
```

https://stackoverflow.com/a/60041791

leogedler: @gregjopa got it, make sense the explanation why is better use to use a const.

I have pushed the suggested change.

gregjopa (approved):
Thanks @leogedler! 💯

PR #121, opened 2021-06-15
feat(storybook): add support for overriding js sdk query params

gregjopa:
This PR updates the storybook demo site to pass query string params through to the JS SDK. 

For example, when running locally and you can pass a custom `client-id` and `components` list in the query string and the JS SDK will apply that:
http://localhost:6006/?path=/story/example-paypalbuttons--default&client-id=sb&components=buttons,messages,hosted-fields

Screen Shot 2021-06-15 at 11 14 54 AM


elizabethmv (approved):
nice!!

Screen Shot 2021-06-15 at 11 22 36 AM

PR #127, opened 2021-07-05
Upgrade Storybook stories to use TypeScript

### LINKED ISSUE #116: [Feature] Use TypeScript for Storybook Examples
### 🚀 Feature Proposal

Currently the Storybook examples are in JavaScript. Let's migrate them over to TypeScript files! It will be a good way to ensure the StoryBook examples are type safe.

### Motivation

Let's go all in on TypeScript!

gregjopa:
This PR upgrades the Storybook examples to use TypeScript. This way there are TypeScript examples of how to type `createOrder` and other props. I also found a couple type bugs that are fixed by this PR.

Resolves #116


.storybook/preview.js
```diff
@@ -1,7 +1,6 @@
 export const parameters = {
     actions: { argTypesRegex: "^on[A-Z].*" },
     docs: {
-        inlineStories: false,
```
gregjopa, line 4: I'm using the `data-namespace` feature of the JS SDK script to support using multiple JS SDK scripts on the same page. This way we don't have storybook wrapping iframes around each component.

Issue #125, opened 2021-07-02
TypeScript declaration files don’t type check


geigerzaehler:
### 🐞 Describe the Bug

The TypeScript declaration files don’t type check when [`skipLibCheck`](https://www.typescriptlang.org/tsconfig#skipLibCheck) is set to `false`.

### 🔬 Minimal Reproduction

```bash
yarn init
yarn add typescript @paypal/react-paypal-js
yarn run tsc --init
echo 'import "@paypal/react-paypal-js"' > index.ts
yarn run tsc --skipLibCheck false
```

### 😕 Actual Behavior

Typescript shows the following error

```plain
node_modules/@paypal/react-paypal-js/dist/index.d.ts:5:25 - error TS7016: Could not find a declaration file for module '@paypal/sdk-constants/dist/module'. '/home/thomas/code/play/paypal-bug/node_modules/@paypal/sdk-constants/dist/module/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/paypal__sdk-constants` if it exists or add a new declaration (.d.ts) file containing `declare module '@paypal/sdk-constants/dist/module';`

5 export { FUNDING } from "@paypal/sdk-constants/dist/module";
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
### 🤔 Expected Behavior

Typescript should not show any errors

### 🌍 Environment

* @paypal/react-paypal-js@7.1.0
* typescript@4.3.5

gregjopa:
Hey @geigerzaehler thanks for letting us know about this type bug.

We are automatically generating the TypeScript declarations for this library: https://github.com/paypal/react-paypal-js/blob/main/tsconfig.json#L3. And that process doesn't seem to automatically include the type declarations for this "@paypal/sdk-constants/dist/module" dependency (https://github.com/paypal/react-paypal-js/blob/main/src/types.d.ts). Any suggestions on how to best fix?

PR #129, opened 2021-07-10
fix: type declarations don’t depend on internal typings

geigerzaehler:
We prevent typescript from generating declarations that reference the `@paypal/sdk-constants` package. This package does not provide types, instead we use our own type declarations in `./src/types.d.ts`. These declarations are not available to users of the library.

Fixes https://github.com/paypal/react-paypal-js/issues/125

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,4 +3,10 @@ export * from "./components/PayPalMarks";
 export * from "./components/PayPalMessages";
 export * from "./ScriptContext";
 
-export { FUNDING } from "@paypal/sdk-constants/dist/module";
+import * as constants from "@paypal/sdk-constants/dist/module";
+
+// We do not re-export `FUNDING` from the `sdk-constants` module
+// directly because it has no type definitions.
+//
+// See https://github.com/paypal/react-paypal-js/issues/125
+export const FUNDING: Record<string, string> = constants.FUNDING;


src/index.ts
```diff
@@ -3,4 +3,10 @@ export * from "./components/PayPalMarks";
 export * from "./components/PayPalMessages";
 export * from "./ScriptContext";
 
-export { FUNDING } from "@paypal/sdk-constants/dist/module";
+import * as constants from "@paypal/sdk-constants/dist/module";
+
+// We do not re-export `FUDING` from the `sdk-constants` module
```
gregjopa, line 8:
```suggestion
// We do not re-export `FUNDING` from the `sdk-constants` module
```

gregjopa:
Awesome work @geigerzaehler! 💯 🥇 👍 

I tested locally and the generated types no longer reference sdk-constants. Here's what I get for index.d.ts:
```ts
export * from "./components/PayPalButtons";
export * from "./components/PayPalMarks";
export * from "./components/PayPalMessages";
export * from "./ScriptContext";
export declare const FUNDING: Record<string, string>;
``` 

PR looks great! I appreciate you leaving a comment in the code explaining what's going on. There's one change request for a typo fix. After that's fixed, I'll merge it up.

gregjopa (approved):

PR #130, opened 2021-07-10
feat: more robust package publishing

geigerzaehler:
* We add the [`prepack` script][prepack] to take care of building the distribution. This removes the extra `build` command from `./scripts/publish.js` and makes it possible to use this package as a git dependency.
* We remove the `dist` folder before packing to avoid including any unnecessary artifacts.
* We explicitly generate type declarations with `tsc`. The rollup plugin only generates type declarations if `output.dir` instead of `output.file` is specified.

[prepack]: https://docs.npmjs.com/cli/v7/using-npm/scripts


gregjopa (approved):
Great improvement! thanks @geigerzaehler 💯 🥇 👍

PR #135, opened 2021-07-28
fix: avoid reloading the script when options have not changed

gregjopa:
## Problem

Currently remounting the `<PayPalScriptProvider>` component always reloads the PayPal JS SDK even when the options have not changed. Reloading the SDK script will destroy and reload components like the Buttons. This is a performance problem and we should avoid unnecessary reloads like this. A test has a been added to cover this use case. It currently fails in the main branch and passes on this feature branch.

## Solution

This PR fixes this bug by generating a hash based on the `<PayPalScriptProvider>` options prop and assigning that to the data-react-paypal-script-id attribute. The paypal-js `loadScript()` function will avoid unnecessary reloads and will reuse the existing script in the DOM as long as the same exact options are passed in to it. Previously we were generating a random number for data-react-paypal-script-id. Now it is deterministic based on the options.


src/ScriptContext.tsx
```diff
@@ -76,8 +82,25 @@ function scriptReducer(state: ScriptContextState, action: ScriptReducerAction) {
     }
 }
 
-function getNewScriptID() {
-    return `react-paypal-js-${Math.random().toString(36).substring(7)}`;
+function getHashFromString(str: string): string {
```
westeezy, line 85: This seems more utility based so maybe this could be in another file. Any reason you are going with this hash implementation specifically? Just curious as we could use something off the shelf as well, which has less cognitive overhead. If this is a has method we want to keep I think a quick function doc block above it and its own file would be really nice for readability.

src/ScriptContext.test.js
```diff
@@ -105,6 +105,37 @@ describe("<PayPalScriptProvider />", () => {
...
+                <TestComponent />
+            </PayPalScriptProvider>
+        );
+
+        await waitFor(() => expect(state.isResolved).toBeTruthy());
+
+        const firstLoadScriptCall = loadScript.mock.calls[0][0];
+        const secondLoadScriptCall = loadScript.mock.calls[1][0];
+
+        expect(firstLoadScriptCall).toEqual(secondLoadScriptCall);
```
westeezy, line 137: great test case really glad to have a test alongside the fix. I want to share where my headspace was as I read this as a newcomer to the codebase. This is hopefully helpful to otherwise reviewing or maybe even could spark an idea or conversation if its interesting to others.
 
**Notes on my thought process as I saw this test:**  
I initially wondered why it even needs to call `loadScript` again if the options haven't changed.  So on my search I found https://github.com/paypal/paypal-js/blob/ccf91267846eb5c1d6b623fb93db1170f640f03f/src/load-script.ts#L26 which looks to prevent from reloading the script. This is great it does do a little pre-work before bailing but nothing major so I didn't see any immediate need to propose further modification based on that.

src/ScriptContext.tsx
```diff
@@ -62,11 +62,17 @@ function scriptReducer(state: ScriptContextState, action: ScriptReducerAction) {
         case "resetOptions":
             // destroy existing script to make sure only one script loads at a time
             destroySDKScript(state.options["data-react-paypal-script-id"]);
```
westeezy, line 64: I don't know the patterns we are using to date but generally speaking you want to keep side effects out of the reducer code. As this is reaching into the dom to remove the script it might be better suited as code to happen within the dispatch of `resetOptions` if we are using an action based approach ala redux or we can quickly wire up a more predictable area for side-effects. This does not need to be resolved for pr approval in my opinion, but I think its a good conversation piece.

gregjopa: That's a great point. We should move this DOM manipulation code outside of the reducer. Let's solve this in a future PR.

westeezy: Sounds great!

westeezy (approved):
thanks for the updates! lgtm!

PR #138, opened 2021-07-30
Break up ScriptProvider into multiple files

borodovisin:
Some minor changes and organization in the project structure


.eslintrc.json
```diff
@@ -4,7 +4,8 @@
     "extends": [
         "eslint:recommended",
         "plugin:@typescript-eslint/recommended",
-        "plugin:react/recommended"
+        "plugin:react/recommended",
+        "plugin:react-hooks/recommended"
```
gregjopa, line 8: Very cool! I didn't know about linting hooks. TIL https://reactjs.org/docs/hooks-rules.html

src/components/PayPalScriptProvider.test.js
```diff
@@ -214,11 +215,13 @@ describe("usePayPalScriptReducer", () => {
 
 function setupTestComponent() {
     const state = {};
-    function TestComponent({ children = null }) {
```
gregjopa, line 217: I'm surprised this change is needed. Can you explain why we need to redeclare like this?

.eslintrc.json
borodovisin: yeah me neither, reading about the React hook rules I saw it...

src/stories/PayPalMarks.stories.tsx
```diff
@@ -6,13 +6,9 @@ import React, {
 } from "react";
 import type { PayPalScriptOptions } from "@paypal/paypal-js/types/script-options";
 
-import {
-    PayPalScriptProvider,
-    PayPalMarks,
-    PayPalButtons,
-    FUNDING,
-} from "../index";
+import { PayPalMarks, PayPalButtons, FUNDING } from "../index";
```
gregjopa, line 9: Can we still get `PayPalScriptProvider` from "../index"?

gregjopa:
Overall this looks like a great refactor for the script provider! Nice work @borodovisin! 💯  I left a few comments/questions.

src/stories/PayPalMarks.stories.tsx
borodovisin: Umm I think we can do that at the end we're exporting it from index, I can fix it...

src/components/PayPalScriptProvider.test.js
borodovisin: To be honest, I do not made any change there I copy the test from file ScriptContext.test.js in main branch, I'll put the latest from main I saw unnecessary declaration of TestComponent variable

src/types/ScriptProvider.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export enum SCRIPT_LOADING_STATE {
+    INITIAL = "initial",
+    PENDING = "pending",
+    REJECTED = "rejected",
+    RESOLVED = "resolved",
+}
+
+export enum DISPATCH_ACTION {
+    LOADING = "setLoadingStatus",
```
westeezy, line 12: As I was reading this I got a bit tripped up with the action type being `LOADING`. I am wondering if it might make sense for readability sake to change this to be something like `SET_LOADING`. I like to view actions similar to intents so their type speaks to the intent of the change vs a namespace name

borodovisin: Completely agree with your comment I put that simple name at first to prove the concept and I forgot to change it. I'll make that change using the same string text to avoid confusions.

src/context/ScriptProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        default: {
+            return state;
+        }
+    }
+}
+
+// Create the React context to use in the script provider component
+export const ScriptContext = createContext<ScriptContextState | undefined>(
+    undefined
```
westeezy, line 83: I generally like to init the reducers with their schema intact vs starting as purely undefined. Though I guess theres not a programatic reason for it but it is helpful I think in terms of readability for me. Let's maybe chat about that offline as a potential thing we can do later.

src/types/ScriptProvider.ts
westeezy: This might be out of scope for the context of file splitting so don't hold the PR on it

src/types/ScriptProvider.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export type InitialState = {
+    options: ReactPayPalScriptOptions;
+    loadingStatus: SCRIPT_LOADING_STATE.INITIAL | SCRIPT_LOADING_STATE.PENDING;
+};
+
+export interface ScriptContextState {
+    options: ReactPayPalScriptOptions;
+    loadingStatus: SCRIPT_LOADING_STATE;
+    dispatch: React.Dispatch<ScriptReducerAction> | null;
```
westeezy, line 40: In what case would this be null?

src/context/ScriptProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+            return {
+                ...state,
+                loadingStatus: action.value as SCRIPT_LOADING_STATE,
+            };
+        case DISPATCH_ACTION.RESET_OPTIONS:
+            // destroy existing script to make sure only one script loads at a time
+            destroySDKScript(state.options[SCRIPT_ID]);
+            // exclude the old data-react-paypal-script-id value from the hash generated by getScriptID()
+            delete (action.value as PayPalScriptOptions)[
+                "data-react-paypal-script-id"
```
gregjopa, line 61:
```suggestion
                SCRIPT_ID
```

src/stories/PayPalScriptProvider.stories.tsx
```diff
@@ -1,8 +1,11 @@
 import React, { FunctionComponent, ReactElement } from "react";
 import type { PayPalScriptOptions } from "@paypal/paypal-js/types/script-options";
 
-import { PayPalScriptProvider, usePayPalScriptReducer } from "../index";
+// import { usePayPalScriptReducer } from "../index";
```
gregjopa, line 4: For storybook, I'd really like us to import every public component from "../index". That way it imitates real integrations with what they can import from the public react-paypal-js module.

src/types/ScriptProvider.ts
westeezy: thanks! sorry it looks like we commented around the same time for those second comments so feel free to make the update of course!

.eslintrc.json
westeezy: great catch!

src/types/ScriptProvider.ts
borodovisin: In my understanding we're dealing with the component state inside a context to use it in all the component tree. Previous version we create two React context. The first one to store the state and the second one to store the dispatch function after we call the useReducer inside the old ScriptProvider functional component. I merged those context in only one. Answering your questions will be null when we call React.createContext(), because will don't have the dispatch at that moment in the src/context/ScriptProvider.ts. We create the initial state inside the PayPalScriptProvider, in the mounting process we call the useReducer and get the reference to the dispatch function and set it in the context provider. Not sure if we can enhance this behavior. Any thoughts are welcome...

src/context/ScriptProvider.ts
borodovisin: Here I made a change (ScriptContextState | null) to initialize the context as null undefined at the end is something not initialized. Probably as you mention we can go directly with the ScriptContextState, not sure at this moment how to do that, because the initial state depends on the ScriptProvider dynamic props, some idea I have is make the ScriptContextState keys optional, but I think that can be a bad idea for static type checking.

src/stories/PayPalScriptProvider.stories.tsx
borodovisin: Sounds good, @gregjopa, I never worked with stories in the past, any other comment or suggestion is welcome...

src/types/ScriptProvider.ts
westeezy: ah okay yeah that sounds totally fine to me. Thanks so much for explaining!

src/context/ScriptProvider.ts
westeezy: Lets leave this as is for now but I have an idea we can discuss in the near future. I think you bring up some good points too so thank you so much for taking a peek! We can chat more and do the work in another PR if we come up with a good plan

src/stories/PayPalScriptProvider.stories.tsx
borodovisin: Done and pushed...

src/components/PayPalScriptProvider.tsx
```diff
@@ -0,0 +1,73 @@
+import React from "react";
```
gregjopa, line 1: Let's import these in so we don't have to prefix with `React.`
```suggestion
import React, { useEffect, useReducer, FC } from "react";
```

src/components/PayPalScriptProvider.tsx
```diff
@@ -0,0 +1,73 @@
...
+                    dispatch({
+                        type: DISPATCH_ACTION.LOADING_STATUS,
+                        value: SCRIPT_LOADING_STATE.REJECTED,
+                    });
+                }
+            });
+        return () => {
+            isSubscribed = false;
+        };
+    }, [options, deferLoading, state.loadingStatus]);
```
gregjopa, line 66: In all the `useEffect` calls, we are purposely excluding some dependencies to avoid re-rendering. Let's add in this comment to each `useEffect` call so eslint does not warn about this.
```suggestion
    // eslint-disable-next-line react-hooks/exhaustive-deps
    }, [options, deferLoading, state.loadingStatus]);
```

gregjopa (approved):
Awesome work @borodovisin! 💯 🥇 👍

src/components/PayPalScriptProvider.tsx
borodovisin: Here I can change the dependency to state.options if you agree, inside the useEffect we aren't using the options object we received in the props of the component

gregjopa: Ahh good point. Let’s update this one

src/components/PayPalScriptProvider.tsx
borodovisin: I'll need to spend some time here, I facing problems with that suggestion. If I remove the React I have problem with lint and typescript. posible solutions are: in .eslintrc.json under rules put this: "react/react-in-jsx-scope": "off" and in .tsconfig.json change compilerOptions.jsx to "react-jsx". But after do that all brakes and I can not build any more the project. here an example of the error:

Screen Shot 2021-07-30 at 5 23 24 PM

borodovisin: Never mind missing extension in nodeResolve() in roolup.config.js. Right config is:
`nodeResolve({"extensions": ".jsx"})`

package.json
```diff
@@ -80,6 +80,7 @@
         "react": "^16.14.0",
         "react-dom": "^16.14.0",
         "react-is": "^17.0.2",
+        "react-test-renderer": "^17.0.2",
```
gregjopa, line 83: Can we uninstall this dependency? It does not seem to be used anywhere
```suggestion
```

src/hooks/ScriptProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+] {
+    const scriptContext = useContext(ScriptContext);
+    if (scriptContext == null || scriptContext?.dispatch == null) {
+        throw new Error(
+            "usePayPalScriptReducer must be used within a PayPalScriptProvider"
+        );
+    }
+
+    const derivedStatusContext = {
+        ...scriptContext!,
```
gregjopa, line 28: Please remove all `!`. They should not be needed since we have the null check on line 21.
```suggestion
        ...scriptContext,
```

src/hooks/ScriptProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+        isInitial:
+            scriptContext!.loadingStatus === SCRIPT_LOADING_STATE.INITIAL,
```
gregjopa, line 30:
```suggestion
            scriptContext.loadingStatus === SCRIPT_LOADING_STATE.INITIAL,
```

src/hooks/ScriptProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+        isPending:
+            scriptContext!.loadingStatus === SCRIPT_LOADING_STATE.PENDING,
```
gregjopa, line 32:
```suggestion
            scriptContext.loadingStatus === SCRIPT_LOADING_STATE.PENDING,
```

src/hooks/ScriptProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+        isResolved:
+            scriptContext!.loadingStatus === SCRIPT_LOADING_STATE.RESOLVED,
```
gregjopa, line 34:
```suggestion
            scriptContext.loadingStatus === SCRIPT_LOADING_STATE.RESOLVED,
```

src/hooks/ScriptProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+        isRejected:
+            scriptContext!.loadingStatus === SCRIPT_LOADING_STATE.REJECTED,
```
gregjopa, line 36:
```suggestion
            scriptContext.loadingStatus === SCRIPT_LOADING_STATE.REJECTED,
```

src/hooks/ScriptProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+    };
+
+    return [derivedStatusContext, scriptContext!.dispatch!];
```
gregjopa, line 39:
```suggestion
    return [derivedStatusContext, scriptContext.dispatch];
```

src/stories/PayPalScriptProvider.stories.tsx
```diff
@@ -1,8 +1,14 @@
-import React, { FunctionComponent, ReactElement } from "react";
+import { FunctionComponent, ReactElement } from "react";
 import type { PayPalScriptOptions } from "@paypal/paypal-js/types/script-options";
 
-import { PayPalScriptProvider, usePayPalScriptReducer } from "../index";
+// import { usePayPalScriptReducer } from "../index";
```
gregjopa, line 4: We can remove the commented out line
```suggestion
```

src/types/ScriptProvider.ts
```diff
@@ -0,0 +1,55 @@
...
+export enum DISPATCH_ACTION {
+    LOADING_STATUS = "setLoadingStatus",
+    RESET_OPTIONS = "resetOptions",
+}
+
+export interface ReactPayPalScriptOptions extends PayPalScriptOptions {
+    [SCRIPT_ID]: string;
+}
+
+export interface ReactPayPalScriptOptions extends PayPalScriptOptions {
```
gregjopa, line 20: This one is a duplicate and can be removed
```suggestion
```

babel.config.js
```diff
@@ -4,7 +4,7 @@ module.exports = (api) => {
     if (isTest) {
         return {
             presets: [
-                ["@babel/preset-react"],
+                ["@babel/preset-react", { runtime: "automatic" }],
```
gregjopa, line 7: I was testing these build changes locally and I'm getting some errors around the jsx runtime. Please remove these babel/tsconfig changes from this PR. We can add them in a separate PR after discussing more. They should not be needed for this PR. 
```suggestion
                ["@babel/preset-react"],
```

src/components/PayPalScriptProvider.tsx
gregjopa: I think there was some confusion here. Sorry about that. We want to keep React as an import. I would like to revert the build changes and import React in along with the other things from React like so:

```js
import React, {
    useEffect,
    useRef,
    useState,
    FunctionComponent,
    ReactElement,
} from "react";
```

westeezy: side note if we update react and the tooling around it we won't need the `React` import anymore. Let's peek into that soon :D

babel.config.js
westeezy: yeah this is an interesting one. we should update the presets and react in sync at a later date

westeezy:
I hate to be a bother but can we mark updated changes as resolved? I was just having a hard time skimming through a minute ago. I'm totally okay if we don't want to as well but figured I'd ask in case we wanted to start doing that as a practice.

gregjopa:
> I hate to be a bother but can we mark updated changes as resolved? I was just having a hard time skimming through a minute ago. I'm totally okay if we don't want to as well but figured I'd ask in case we wanted to start doing that as a practice.

Yes we should. Good point.

package.json
borodovisin: Sure!!!

src/components/PayPalScriptProvider.tsx
borodovisin: I already solve the problems with the React import. I can test it again and check is something is missing. I'm able to build and run test in the project after removing the React and start using the jsx transform after react version 17.

babel.config.js
borodovisin: Hi @gregjopa, please can you explain me a little the problems you're facing because I had the same issue with jsx and runtime error in the console trying to build de project or run the test cases...

borodovisin: Also I have a question, I my personal case I prefer *.yml over json to make configuration can I change that. It is not a big deal is just a preference and for me is more readable yaml files.

src/components/PayPalScriptProvider.tsx
borodovisin: Can we update the package to only allows REact version >=17?

borodovisin: I prefer to go with the new React transform jsx. I can make a suggestion this is my principal reason:

Screen Shot 2021-07-31 at 5 46 09 PM

Screen Shot 2021-07-31 at 5 46 27 PM

The size in the imports

gregjopa: > I already solve the problems with the React import. I can test it again and check is something is missing. I'm able to build and run test in the project after removing the React and start using the jsx transform after react version 17.

I see two problems when testing locally:

1. VS Code shows type errors in index.ts. There's a VS Code bug related to react-jsx showing these as type errors. There are not actually type errors when running `npm run typecheck` but it's still confusing and we should avoid for now.

Screen Shot 2021-08-01 at 11 34 28 AM

2. The task `npm run build` throws a new error related to this change:

Screen Shot 2021-08-01 at 11 35 09 AM

> Can we update the package to only allows React version >=17?

We currently support react@16.3 or greater. I do not want to change that (at least not in this PR). Also are you measuring bundle size using the dist files? We have react as a peer dependency so its not included in the actual bundles.

babel.config.js
gregjopa: > Also I have a question, I my personal case I prefer *.yml over json to make configuration can I change that. It is not a big deal is just a preference and for me is more readable yaml files.

Let's not change any config files. I'm open to new ideas but I would like to discuss in a new issue and/or offline first.

Please keep this PR scoped to it's original purpose of breaking up the ScriptProvider component into multiple files. The new react-hooks eslint plugin is great so lets keep that too. Adding any new/additional features to this PR makes it difficult for @westeezy and I can review/approve. I would like to get this merged since it's great work and the original problem is solved. Future changes can be separate PRs. It's best practice to keep PR's feature-focused.

borodovisin: Yeah, complete agree with you, isn't good for anyone start tracking a bunch of files.

gregjopa (approved):
Awesome work @borodovisin! Thanks for the contribution 🚀

PR #139, opened 2021-08-12
Dtppsdk 239

borodovisin:
New React component <BraintreePayPalButton /> to enable Braintree integration in the project.

The new component is using the PayPalButtons component, it is somehow a wrapper component to override the createOrder and onApprove function using the Braintree paypalCheckout


borodovisin:
Fixed test coverage in next PR and pass all checks

PR #141, opened 2021-08-17
DTPPSDK-275: Fix hashStr function and unit test coverage

borodovisin:
Fixed overflow problem with the hashStr function and unit test coverage


src/utils.test.js
```diff
@@ -0,0 +1,24 @@
+import { hashStr } from "./utils";
+
+describe("hashStr", () => {
+    test("should match the hash from the argument string", () => {
+        expect(hashStr("react")).toMatchSnapshot();
```
westeezy, line 5: Sorry to be a bit picky on this comment, but generally speaking for tests like this I'd almost rather just see the expected string inline vs in a snapshot. I find when everything is contained in one file its easier to quickly see expected behavior. Given that these are shorter strings I think inlining would be easier than snapshots, but I am totally open to whatever folks feel is best!

borodovisin: Sure I can use the toMatchInlineSnapshot because I'm calling a hash function and don't know from the top of my head the result

src/utils.test.js
```diff
@@ -0,0 +1,30 @@
+import { hashStr } from "./utils";
+
+describe("hashStr", () => {
+    test("should match the hash from the argument string", () => {
+        expect(hashStr("react")).toMatchInlineSnapshot(`"xxhjw"`);
```
gregjopa, line 5: TIL about Inline Snapshots. Very cool!

https://jestjs.io/docs/snapshot-testing#inline-snapshots

gregjopa (approved):
Nice work @borodovisin!

westeezy (approved):
lgtm

src/utils.test.js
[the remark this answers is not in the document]
borodovisin: Yeah, that is right, the only think I do not create the string, that was generate from the jest library, because that I not use the toEqual. To make that I need to run the function externally and copy the result.

PR #154, opened 2021-09-01
PayPalMarks allows re-render on fundingSource changes

borodovisin:
The PayPalMarks component doesn't re-render on fundingSource props changes. The component has the fundingSource as a dependency in the useEffect but was disabled because an if statement inside the useEffect hook.

With this PR we fix that problem and allows re-render the PayPalMarks component when any useEffect dependency changes.

### DIFF
--- a/src/components/PayPalMarks.test.js
+++ b/src/components/PayPalMarks.test.js
@@ -139,24 +139,21 @@ describe("<PayPalMarks />", () => {
         const spyConsoleError = jest
             .spyOn(console, "error")
             .mockImplementation();
+        const mockRender = jest
+            .fn()
+            .mockRejectedValue(new Error("Unknown error"));
         window.paypal.Marks = () => ({
             isEligible: jest.fn().mockReturnValue(true),
-            render: jest.fn((element) => {
-                element.removeAttribute("class");
-                return Promise.reject(new Error("Unknow error"));
-            }),
+            render: mockRender,
         });
-        const { container } = render(
+
+        render(
             <PayPalScriptProvider options={{ "client-id": "test" }}>
                 <PayPalMarks className="test-class" />
             </PayPalScriptProvider>
         );
 
-        await waitFor(() =>
-            expect(
-                container.querySelector(".test-class") instanceof HTMLDivElement
-            ).toBeFalsy()
-        );
+        await waitFor(() => expect(mockRender).toBeCalled());
         spyConsoleError.mockRestore();
     });
 
@@ -179,34 +176,32 @@ describe("<PayPalMarks />", () => {
     });
 
     test("should rerender the component when the funding source change", async () => {
+        const mockRender = jest.fn((element) => {
+            const markElement = document.createElement("div");
+
+            markElement.setAttribute("id", "children-div");
+            element.append(markElement);
+            return Promise.resolve();
+        });
         window.paypal.Marks = () => ({
             isEligible: jest.fn().mockReturnValue(true),
-            render: (element) => {
-                const markElement = document.createElement("div");
-
-                markElement.setAttribute("id", "children-div");
-                element.append(markElement);
-                return Promise.resolve();
-            },
+            render: mockRender,
         });
 
-        const { container, rerender } = render(
+        const { rerender } = render(
             <PayPalScriptProvider options={{ "client-id": "test" }}>
                 <PayPalMarks fundingSource="paypal" />
             </PayPalScriptProvider>
         );
 
-        await waitFor(() =>
-            expect(
-                container.querySelector("#children-div") instanceof
-                    HTMLDivElement
-            ).toBeTruthy()
-        );
+        await waitFor(() => expect(mockRender).toBeCalledTimes(1));
 
         rerender(
             <PayPalScriptProvider options={{ "client-id": "test" }}>
                 <PayPalMarks fundingSource="card" />
             </PayPalScriptProvider>
         );
+
+        await waitFor(() => expect(mockRender).toBeCalledTimes(2));
     });
 });
--- a/src/components/PayPalMarks.test.js
+++ b/src/components/PayPalMarks.test.js
@@ -160,31 +160,24 @@ describe("<PayPalMarks />", () => {
         spyConsoleError.mockRestore();
     });
 
-    test("should not render component when is not eligible", async () => {
+    test("should not render component when ineligible", async () => {
+        const mockIsEligible = jest.fn().mockReturnValue(false);
+        const mockRender = jest.fn().mockResolvedValue();
+
         // Make Marks not eligible
         window.paypal.Marks = () => ({
-            isEligible: jest.fn().mockReturnValue(false),
-            render: (element) => {
-                const spanElement = document.createElement("span");
-
-                spanElement.setAttribute("id", "span-element");
-                element.append(spanElement);
-                return Promise.resolve();
-            },
+            isEligible: mockIsEligible,
+            render: mockRender,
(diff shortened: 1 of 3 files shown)


westeezy (commented):
Can you add a test for this? Otherwise looks good to me

src/components/PayPalMarks.tsx
```diff
@@ -40,19 +36,40 @@ export const PayPalMarks: FunctionComponent<PayPalMarksComponentProps> = ({
...
 
-        // don't rerender when already rendered
-        if (mark.current !== null) {
+        // only render the mark when eligible
+        if ((mark != null && mark.isEligible()) === false || current == null)
             return;
-        }
+        // Remove any children before
+        if (current !== null && current.firstChild != null)
+            /* istanbul ignore next */ // Can't find a solution to test the next line
```
borodovisin, line 52: @gregjopa and @westeezy Please, I try to cover this line in the test cases, but not find a way to do it using the --env=jsdom. Using the render from testing-library all the time the ref.current.firstChild is null. I put a comment to avoid problem with the test coverage. If do you have any idea to solve this, please let me know.

src/components/PayPalMarks.test.js
```diff
@@ -11,6 +11,15 @@ jest.mock("@paypal/paypal-js", () => ({
     loadScript: jest.fn(),
 }));
 
+jest.mock("react", () => ({
```
gregjopa, line 14: Usually we use `document.body.innerHTML = "";` to set the state of the DOM for unit tests. Can we setup the test using `innerHTML` instead of mocking React `useRef`? Let me know if there's a problem with that approach.

src/context/scriptProviderContext.ts
```diff
@@ -35,8 +35,7 @@ export function destroySDKScript(reactPayPalScriptID: string): void {
         `script[${SCRIPT_ID}="${reactPayPalScriptID}"]`
     );
 
-    if (scriptNode != null && scriptNode.parentNode)
-        scriptNode.parentNode.removeChild(scriptNode);
+    if (scriptNode?.parentNode) scriptNode.parentNode.removeChild(scriptNode);
```
westeezy, line 38:
```suggestion
    if (scriptNode?.parentNode) {
        scriptNode.parentNode.removeChild(scriptNode);
    }
```

very minor nitpick so feel free to reject this suggestion. just throwing it out there as i prefer all conditionals in the bracketed syntax for easier readability. the minifier can simplifier it down later

westeezy (approved):
lgtm thanks for adding the tests as well 💪

src/components/PayPalMarks.test.js
borodovisin: Never mind that was something I start testing that piece of code is not needed. sorry about that.

src/context/scriptProviderContext.ts
borodovisin: Not a problem at all. I'll change to keep the bracketed syntax, I guess is the standard.

src/components/PayPalMarks.test.js
```diff
@@ -134,4 +134,88 @@ describe("<PayPalMarks />", () => {
...
+
+        await waitFor(() =>
+            expect(
+                container.querySelector(".test-class") instanceof HTMLDivElement
+            ).toBeFalsy()
+        );
+        spyConsoleError.mockRestore();
+    });
+
+    test("should not render component when is not eligible", async () => {
```
gregjopa, line 163:
```suggestion
    test("should not render component when ineligible", async () => {
```

src/components/PayPalMarks.test.js
```diff
@@ -134,4 +134,88 @@ describe("<PayPalMarks />", () => {
...
+        // Make Marks not eligible
+        window.paypal.Marks = () => ({
+            isEligible: jest.fn().mockReturnValue(false),
+            render: (element) => {
```
gregjopa, line 167: Do we need to mock render for this test? If we do, could we use:

```js
jest.fn().mockResolvedValue()
```

src/components/PayPalMarks.test.js
```diff
@@ -134,4 +134,88 @@ describe("<PayPalMarks />", () => {
...
+
+                spanElement.setAttribute("id", "span-element");
+                element.append(spanElement);
+                return Promise.resolve();
+            },
+        });
+
+        const { container } = render(
+            <PayPalScriptProvider options={{ "client-id": "test" }}>
+                <PayPalMarks />
```
gregjopa, line 178: Can we pass in a `className` here and then use that as the selector for the assertion?

gregjopa (approved):
Looks great! I left a couple comments around the tests

src/components/PayPalMarks.test.js
```diff
@@ -134,4 +134,81 @@ describe("<PayPalMarks />", () => {
...
+            ).toBeFalsy()
+        );
+        spyConsoleError.mockRestore();
+    });
+
+    test("should not render component when ineligible", async () => {
+        const mockIsEligible = jest.fn().mockReturnValue(false);
+        const mockRender = jest.fn().mockResolvedValue();
+
+        // Make Marks not eligible
```
gregjopa, line 167: lets remove the comment. I think the new variables are self describing.
```suggestion
```

src/components/PayPalMarks.test.js
```diff
@@ -134,4 +134,81 @@ describe("<PayPalMarks />", () => {
...
+                element.removeAttribute("class");
+                return Promise.reject(new Error("Unknow error"));
+            }),
+        });
+        const { container } = render(
+            <PayPalScriptProvider options={{ "client-id": "test" }}>
+                <PayPalMarks className="test-class" />
+            </PayPalScriptProvider>
+        );
+
```
gregjopa, line 154: Can we assert that render() got called?

src/components/PayPalMarks.test.js
```diff
@@ -134,4 +134,81 @@ describe("<PayPalMarks />", () => {
...
+        await waitFor(() => expect(mockIsEligible).toBeCalled());
+        expect(mockRender).not.toBeCalled();
+    });
+
+    test("should rerender the component when the funding source change", async () => {
+        window.paypal.Marks = () => ({
+            isEligible: jest.fn().mockReturnValue(true),
+            render: (element) => {
```
gregjopa, line 186: Can we assert that render gets called after the first render and then again after rerender?

src/components/PayPalMarks.test.js
```diff
@@ -139,24 +139,21 @@ describe("<PayPalMarks />", () => {
...
         });
-        const { container } = render(
+
+        render(
             <PayPalScriptProvider options={{ "client-id": "test" }}>
                 <PayPalMarks className="test-class" />
             </PayPalScriptProvider>
         );
 
-        await waitFor(() =>
```
gregjopa: Let's keep the DOM assertion as well. Hoping this will work.

```js
await waitFor(() => expect(mockRender).toBeCalled());
expect(
  container.querySelector(".test-class") instanceof HTMLDivElement
).toBeFalsy()
```

borodovisin: I think this won't work because the PayPalMarks component will keep all the time the div container, that means the assertion above will fail all the time. The render process fail but doesn't remove the div containing the class `test-class`.

PayPalMarks always return this: `<div ref={markContainerRef} className={className} />` never modify this nodeElement.

gregjopa: Good point. Ignore this one.

PR #155, opened 2021-09-13
Support PayPalMarks children

borodovisin:
With this PR we enable children in the PayPalMarks component. To allows the creation of custom components when the marks   is ineligible ins the same way we support in PayPalButtons.

Now we can use this code below: 
```jsx
<PayPalMarks className="mark-container">
    <div className="ineligible"></div>
</PayPalMarks>
```

When Marks is ineligible will render the children component.

### DIFF
--- a/src/components/PayPalMarks.tsx
+++ b/src/components/PayPalMarks.tsx
@@ -1,4 +1,12 @@
-import React, { useEffect, useRef, useState, FunctionComponent } from "react";
+import React, {
+    useEffect,
+    useRef,
+    useState,
+    FunctionComponent,
+    ReactElement,
+    ReactPortal,
+    ReactFragment,
+} from "react";
 import { usePayPalScriptReducer } from "../hooks/scriptProviderHooks";
 import { getPayPalWindowNamespace } from "../utils";
 import { DEFAULT_PAYPAL_NAMESPACE } from "../constants";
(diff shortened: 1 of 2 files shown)


src/components/PayPalMarks.tsx
```diff
@@ -89,7 +102,17 @@ export const PayPalMarks: FunctionComponent<PayPalMarksComponentProps> = ({
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [isResolved, markProps.fundingSource]);
 
-    return <div ref={markContainerRef} className={className} />;
+    return (
+        <>
+            {isEligible ? (
+                <div ref={markContainerRef} className={className}>
+                    {children}
```
gregjopa, line 109: I think we should only render the `children` when `isEligible` is false.
```suggestion
```

borodovisin: Yeah, my mistake here, I forgot to remove it...

src/components/PayPalMarks.tsx
```diff
@@ -12,6 +20,7 @@ export interface PayPalMarksComponentProps extends PayPalMarksComponentOptions {
      * Pass a css class to the div container.
      */
     className?: string;
+    children?: ReactElement | ReactPortal | ReactFragment;
```
gregjopa, line 23: In a separate PR, can you update the Buttons `children` prop to have the same type signature?

gregjopa (approved):

PR #157, opened 2021-09-14
Same signature on all the children components

borodovisin:
In this PR we unified the children signature type on all the component [PayPalButtons, PayPalBraintreeButtons, PayPalMarks].


src/types/scriptProviderTypes.ts
```diff
@@ -45,3 +46,5 @@ export interface ScriptProviderProps {
     children?: React.ReactNode;
     deferLoading?: boolean;
 }
+
+export type PayPalProviderChildren = ReactElement | ReactPortal | ReactFragment;
```
gregjopa, line 50: I'd prefer to keep the name general.
```suggestion
export type Children = ReactElement | ReactPortal | ReactFragment;
```

gregjopa (approved):

PR #164, opened 2021-09-23
New PayPal Hosted Fields story

borodovisin:
In this PR we can find the stories for new PayPalHostedFieldsProvider -> PayPalHostedField components.

Now it is only two stories. One with expiration date as a single field and the second with expiration date split in two fields.

Here is missing the postal code fields from the payPalHostedFields SDK. Not sure if it is needed in the stories.


src/components/hostedFields/utils.ts
```diff
@@ -65,9 +69,9 @@ export const decorateHostedFields = (
 
     return {
         ...(hostedFields as PayPalHostedFieldsComponent),
-        close(container) {
+        close(container, selectors = hostedFieldsChildrenSelector) {
             if (container) {
-                container.querySelectorAll("*").forEach((element) => {
+                container.querySelectorAll(selectors).forEach((element) => {
```
borodovisin, line 74: Any comment here is welcome. To clean the previously hosted fields I'm using this selector `iframe, div[style='clear: both;']` to remove from the DOM all the created iframe and the div inserted after the iframe

src/components/hostedFields/utils.ts
```diff
@@ -17,6 +17,10 @@ import type {
     PayPalHostedFieldOptions,
 } from "../../types/payPalHostedFieldTypes";
 
+// This represent the children to remove manually inside the PayPalHostedFieldsProvider
```
gregjopa, line 20:
```suggestion
// Mock children inside the PayPalHostedFieldsProvider to simulate the HostedFields component
```

src/components/hostedFields/utils.ts
```diff
@@ -17,6 +17,10 @@ import type {
...
+// The HostedFields SDK manually insert and iframe and a div for each declared field
```
gregjopa, line 21:
```suggestion
```

PR #166, opened 2021-09-30
Refactor and split PR

borodovisin:
In this PR we can find some refactors about constants variables.
Remove the CHildren type and use the ReactNode instead.
Rename some constants variables
Changes on the context validator


westeezy (approved):
lgtm

PR #173, opened 2021-10-01
refactor(tests): convert test utils to typescript

pedpess:
This PR converts the `utils.test.js` to Typescript and adds `@types/jest` lib to the project. All tests are passing ✅ , but worth checking also the `window.braintree` object that comes from `paypal-js` core lib as `BraintreeNamespace` is not declared there.


src/utils.test.ts
```diff
@@ -9,6 +9,7 @@ describe("getPayPalWindowNamespace", () => {
     beforeAll(() => {
         window.paypal = {
             Buttons: jest.fn(),
+            version: "1.0.0",
```
gregjopa, line 12: This shouldn't be needed.
```suggestion
```

gregjopa:
Hi @pedpess, thanks for sending this PR! 

I'm seeing failures in CI. Can you run `npm run validate` locally to make sure your changes are passing with the current rules and test suite?

gregjopa:
> All tests are passing ✅ , but worth checking also the window.braintree object that comes from paypal-js core lib as BraintreeNamespace is not declared there, so your tests will fail.

> Here's the line:
https://github.com/paypal/paypal-js/blob/cdc5608bfb827a91e21c18039fc5951502207208/types/index.d.ts#L52

All the Braintree types are declared in this react-paypal-js repo. The paypal-js repo is only for the PayPal JS SDK. Here's where we define the types for the global braintree namespace that gets set by the braintree-web sdk: https://github.com/paypal/react-paypal-js/blob/main/src/utils.ts#L31-L33

davidalpert:
I tried running `npm ci` on this branch and got the following error:
```
npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
```

When you installed the `@types/jest` dependency did you use `npm install`? that usually updates `package-lock.json` also...

gregjopa:
I went ahead and added `@types/jest` to the main branch in this commit: https://github.com/paypal/react-paypal-js/commit/dbeecd9a97e0401db70af43b9d6237026651feac. @pedpess please rebase this branch with main. Thank you!

pedpess:
Thanks, guys, I will continue working on that when I'm back home from the office 👍🏻

src/utils.test.ts
pedpess: Yes, `version` is missing here... `paypal` window object is optional, but `version` field inside of it mandatory from the `PayPalNamespace` namespace definition 🤔

src/utils.test.ts
```diff
@@ -22,16 +25,17 @@ describe("getPayPalWindowNamespace", () => {
...
 describe("getBraintreeWindowNamespace", () => {
+    const mockBraintreeNamespace = mock<BraintreeNamespace>();
+
     beforeAll(() => {
-        window.braintree = {
-            createPayment: jest.fn(),
-            client: jest.fn(),
-        };
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (window as any).braintree = mockBraintreeNamespace;
```
pedpess, line 32: **thought:**

@gregjopa both tests and validation scripts are passing ✅ . However, I was not happy enough because I needed to cast `any` here. But considering this is non-prod code is it fine?

As mentioned previously. I didn't think anything else wiser other than adding the `BraintreeNamespace` in the type declaration file from `paypal-js` like this:

```typescript
declare global {
    interface Window {
        paypal?: PayPalNamespace;
        braintree?: BraintreeNamespace;
    }
}
```

davidalpert: I wondered about this in #174; I didn't know enough about BraintreeNamespace or how this global.Window interface was used to suggest this but if it is accepted then we could replace the use of `(window as any)` in the affected test class from #174

Issue #170, opened 2021-10-01
 [Hacktoberfest] Covert BraintreePayPalButtons.test.js to TypeScript


gregjopa:
We have a bunch of tests that are still using plain JavaScript even though this library is written in TypeScript.

For this issue, convert the [BraintreePayPalButtons.test.js](
https://github.com/paypal/react-paypal-js/blob/main/src/components/braintree/BraintreePayPalButtons.test.js) to TypeScript.

davidalpert:
I can take this one, but I intend to convert `src/components/braintree/utils.test.js` first, and may need to branch off of `pedpess:refactor/convert-test-utils-to-typescript` to avoid a conflict if this new PR needs `@types/jest` also.

gregjopa:
@davidalpert sounds great. I went ahead and added `@types/jest` to the main branch here: https://github.com/paypal/react-paypal-js/commit/dbeecd9a97e0401db70af43b9d6237026651feac. That way none of the other PRs are stuck waiting for it.

davidalpert:
draft PR is here: https://github.com/paypal/react-paypal-js/pull/174

davidalpert:
@gregjopa it looks like this project is missing a `hacktoberfest` topic at the repo level, as [hacktoberfest](https://hacktoberfest.digitalocean.com/profile) is telling me that if found my PR (probably based on the label) but that the project `react-paypal-js` is not participating.

gregjopa:
> @gregjopa it looks like this project is missing a `hacktoberfest` topic at the repo level, as [hacktoberfest](https://hacktoberfest.digitalocean.com/profile) is telling me that if found my PR (probably based on the label) but that the project `react-paypal-js` is not participating.

Hi @davidalpert, I added the hacktoberfest label. Hopefully that resolves the issue. We do want react-paypal-js participating and the other repos linked from this page: https://github.com/paypal/hacktoberfest

davidalpert:
that did it 🎉 🎃 ⌨️ 

my PR is now showing up as part of a participating repository. thank you.

davidalpert:
with #174 merged I believe that this issue has been resolved 👏

PR #174, opened 2021-10-02
GH-170 convert braintree components to typescript

davidalpert:
converts:

- [x] `src/components/braintree/utils.test.js` -> typescript
- [x] `src/components/braintree/BraintreePayPalButtons.test.js` -> typescript

introduces the `jest-mock-extended` package to clean up mock test data which doesn't impact behavior


gregjopa:
The jest-mock-extended library looks awesome! Great suggestion. Definitely makes mocking easy with TypeScript.

davidalpert:
having a bit more trouble with the setup in the `BraintreePayPalButtons.test.tsx` but I'm making progress.

davidalpert:
think that I have the per-test setup worked out.

now I've got one test case failing on `FUNDING` not defined; looks like I'm tripping over something similar to #125 so I'm looking into why Jest is failing to accept `export` in an imported file.

davidalpert:
think I cracked the last compilation issue. details in that last commit message.

davidalpert:
I needed to use the same node version as the build is using in order to reproduce the validation error (issues resolving the types of react-dom).

How would you feel about adding a `.node-version` file to the root of this repo containing
```
12.22.6
```
to be explicit about which node version contributors should use in order to reproduce the behaviour on the build server?

davidalpert:
that code coverage report looks to be reporting a drop in code coverage due to three new files but those new files do not seem to be added in this PR... 🤔

src/types/scriptProviderTypes.ts
```diff
@@ -43,7 +43,7 @@ export interface ScriptContextDerivedState
 }
 
 export interface ScriptProviderProps {
-    options: PayPalScriptOptions;
+    options?: PayPalScriptOptions;
```
gregjopa, line 46: We need the `options` prop to be required since that's how `client-id` is passed through. The JS SDK will return a 400 status code if client-id is not provided as a query param.

```suggestion
    options: PayPalScriptOptions;
```

src/components/braintree/BraintreePayPalButtons.test.tsx
```diff
@@ -130,7 +146,7 @@ describe("Braintree PayPal button fail in mount process", () => {
                 </PayPalScriptProvider>
             );
         } catch (ex) {
-            errorMessage = ex.message;
+            errorMessage = (ex as Error).message;
```
gregjopa, line 149: Very cool! I've been wondering how to properly cast error objects in TS. This approach is nice and simple.

src/components/braintree/utils.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+        const mockedCheckout = mock<BraintreePayPalCheckout>();
+
+        const buttonProps = decorateActions(
+            buttonPropsWithOnApprove,
+            mockedCheckout
+        );
+
+        if (!buttonProps) {
+            fail("buttonProps is undefined");
```
gregjopa, line 91: I didn't realize that `fail()` was a thing in Jest. Looks like it comes from Jasmine and Jest is built upon that. I don't see `fail()` referenced in the Jest docs though. Can we stick with the existing pattern of throwing an error for failing a test?

```suggestion
            throw new Error("buttonProps is undefined");
```

src/components/braintree/utils.test.ts
```diff
@@ -0,0 +1,102 @@
...
+        } else if (!buttonProps.onApprove) {
+            fail("buttonProps.onApprove is undefined");
```
gregjopa, line 93:
```suggestion
            throw new Error("buttonProps.onApprove is undefined");
```

gregjopa:
@davidalpert this PR is really impressive! 💯 🥇 👍  It will be a great example for the other TypeScript conversion tasks. Thanks for the contribution! I left a few comments. After those are addressed, we should be good to merge.

gregjopa:
> that code coverage report looks to be reporting a drop in code coverage due to three new files but those new files do not seem to be added in this PR... 🤔

My initial guess is this is caused by your fork being out of date with the main branch on upstream. Can you try rebasing with the latest from main? This is not a blocker. I'm happy to merge with the current state of the coverage since those files appear to be unrelated to the changes in the PR.

gregjopa:
> I needed to use the same node version as the build is using in order to reproduce the validation error (issues resolving the types of react-dom).
> 
> How would you feel about adding a `.node-version` file to the root of this repo containing
> 
> ```
> 12.22.6
> ```
> 
> to be explicit about which node version contributors should use in order to reproduce the behaviour on the build server?

This is a good point. Can you create a new issue for enforcing a node version? And we can have the discussion there.

src/types/scriptProviderTypes.ts
davidalpert: as I recall this was needed to make one of the tests compile after converting it to typescript. I can look at that again; if the test that fails to compile is a demonstration of the requirement that you are referring to then I will look further into how to make that test valid without failing the type checking.

davidalpert:
I appreciate the feedback. I will work a bit more with this locally and then once I'm sure that the tests are passing I will rebase onto `upstream/main`

gregjopa:
Feel free to use a `// @ts-expect-error` comment for the ts errors around a missing `options` prop. For some of the tests, we are purposely not passing that it so its safe to ignore those compile errors.

davidalpert:
sweet. I didn't know about the `// @ts-expect-error` directive.

added that to `BraintreePayPalButtons.test.tsx` and now the `validate` script passes locally on my machine (using node `12.22.6`)

davidalpert:
@gregjopa I have rebased onto `upstream/main` and took the opportunity to squash/fix some of those commits.

Most notably the commit which converts `BraintreePayPalButtons.test.js` --> `BraintreePayPalButtons.test.tsx` now includes the entire conversion in a single commit and no longer includes any changes to the code under test 👍 

if you prefer I could separate out the jest config changes needed in `package.json` into a separate commit but that is likely overkill if you plan to merge this shortly; let me know.

src/components/braintree/utils.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+        const mockedCheckout = mock<BraintreePayPalCheckout>();
+
+        const buttonProps = decorateActions(
+            buttonPropsWithCreateOrder,
+            mockedCheckout
+        );
+
+        if (!buttonProps) {
+            fail("buttonProps is undefined");
```
gregjopa, line 58:
```suggestion
            throw new Error("buttonProps is undefined");
```

src/components/braintree/utils.test.ts
```diff
@@ -0,0 +1,102 @@
...
+        } else if (!buttonProps.createOrder) {
+            fail("buttonProps.createOrder is undefined");
```
gregjopa, line 60:
```suggestion
            throw new Error("buttonProps.createOrder is undefined");
```

gregjopa (approved):
Awesome work @davidalpert! 💯 🥇 👍

Issue #171, opened 2021-10-01
[Hacktoberfest] Covert contextValidator.test.js to TypeScript


gregjopa:
We have a bunch of tests that are still using plain JavaScript even though this library is written in TypeScript.

For this issue, convert the [contextValidator.test.js](https://github.com/paypal/react-paypal-js/blob/main/src/hooks/contextValidator.test.js) to TypeScript.

pedpess:
Hey again Greg 👋🏻 , I can take this one too. I will continue working on it in few hours

itsdharmxd:
**@gregjopa**  can you please assign this project

sagarchoudhary96:
@gregjopa  @itsdharmxd if it's not completed can i pick it up ?

itsdharmxd:
> 
> 
> @gregjopa @itsdharmxd if it's not completed can i pick it up ?

sure go for it

sagarchoudhary96:
@gregjopa once file is changed to `.ts` file it will start throwing error for invalid param type passed to the functions based on the type check. what all needs to be changed in the test FIle then ?

sagarchoudhary96:
@gregjopa please review the PR.

PR #178, opened 2021-10-05
refactor(tests): convert contextValidator.test.js to typescript file

sagarchoudhary96:
Related to https://github.com/paypal/react-paypal-js/issues/171


westeezy (approved):
lgtm. thanks for looking into this! the tests definitely do also indicate we can revisit the types for `contextNotEmptyValidator` in the future as we really should see if we can do away with the need to cast to `never` in the implementing code. however because the way it is written this PR is great to get us to TS 🎉

westeezy:
workflows appear stuck on this. will manually validate

Issue #177, opened 2021-10-04
express specific node version for use in local development


davidalpert:
While working on PR #174 I discovered that I needed to use the same node version locally as the build was using in order to reproduce some validation errors seen only when the validate script ran inside github actions.

I propose adding a `.node-version` file to the root of containing

```
12.22.6
```

in order to be explicit about which node version contributors should use for local development.

node version managers like `nodenv` and `nvm` can take advantage of such local hints automatically to help increase consistency between the local development experience and what github actions are doing.

> This is a good point. Can you create a new issue for enforcing a node version? And we can have the discussion there.

_Originally posted by @gregjopa in https://github.com/paypal/react-paypal-js/issues/174#issuecomment-933521236_

PR #180, opened 2021-10-05
chore: pin node version using .nvmrc

wsbrunson:
Addresses: https://github.com/paypal/react-paypal-js/issues/177

Some explanations for the route I took:

## .nvmrc
We went with the .nvmrc format because at PayPal the default Node manager is nvm. I did some research and its not really clear what file format the community is trying to standardize on (`.nvmrc` vs `.node-version` vs package.json's `engines`). 

I've seen others symlinking between a `.nvmrc` and `.node-version` file and there's [this package](https://www.npmjs.com/package/package-node-to-nvmrc) to keep `.nvmrc` and package.json's `engines` synced using pre-commit hooks. Probably not a bad idea to start with `.nvmrc` and see what other Node managers people are using.

## v14.18
With the `.nvmrc` we can be as loose as `v14` or as strict as `v14.18.0`. I figured we don't need to be on the semi-bleeding edge of Node and I couldn't think of a good reason why we'd need to be extra strict with what version we used. I settled on the happy medium of Major + Minor version. Open to other ideas here though.

## Github Actions
For this repo, the only action that specifically calls out node is the `validate` action. It's using a strategy matrix of version 12 and 14. I'm not sure if we've ever talked about what versions of past Node we will support. I'm happy to leave this alone if it's too much discussion or move our `validate`  action to using the pinned Node version.

## Ignore storybook-static output
When testing the scripts we use for this repo for Node v14.18, I noticed that one of the storybook commands outputs a `storybook-static` folder which wasn't being ignored by git or eslint


davidalpert:
I appreciate the thoroughness of this PR description 👏 

picking `nvm` (and thus `.nvmrc`) given that's what paypal uses makes total sense.

the reason I opened #177 is that I was contributing to #174 and ran into a situation where the validate script "worked on my machine" and failed in the PR checks. I wasn't able to reproduce the failure until I matched the node version in the failing PR check.

interesting that the actions are testing against a matrix of node versions. I wonder whether that's necessary, since the project is in typescript.  As I understand it, tsconfig settings let you pick compatibility of the transpiled lib code

gregjopa:
>For this repo, the only action that specifically calls out Node is the validate action. It's using a strategy matrix of version 12 and 14. I'm not sure if we've ever talked about what versions of past Node we will support. I'm happy to leave this alone if it's too much discussion or move our validate action to using the pinned Node version.

Let's remove the matrix functionality. Originally I thought building and testing with multiple Node versions was a good idea. But I don't think it's worth it. Let's drop the matrix functionality and replace it with using this single node version instead.

package.json
```diff
@@ -17,21 +17,22 @@
     "types": "dist/types/index.d.ts",
     "scripts": {
         "build": "rollup --config",
+        "check-node-version": "check-node-version --node $(cat .nvmrc)",
```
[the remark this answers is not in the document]
wsbrunson: Greg and I talked on Slack and decided to create something similar to this this script but for checking the node version using the .nvmrc file: https://github.com/paypal/paypal-js/blob/main/scripts/readme-package-version.js

scripts/check-node-version.js
```diff
@@ -0,0 +1,69 @@
...
+        throw err;
+    }
+
+    printVersions(result);
+    process.exit(result.isSatisfied ? 0 : 1);
+});
+
+// printVersions and printInstalledVersion were copied from:
+// https://github.com/parshap/check-node-version/blob/master/cli.js
+
```
wsbrunson, line 19: Unfortunately when we use `check` from `check-node-version`, we have to implement a lot of the functionality from the CLI ourselves. I went down a few paths to try to avoid this but those solutions had their own problems. To maintain cross-platform compatibility, this is probably the simplest solution.

gregjopa (approved):

scripts/check-node-version.js
wsbrunson: Okay that worked well! We still have the problem with no colors in the output which makes it hard to find what went wrong. I added an error message and used `chalk` to make it stand out, let me know what you think.

Image 2021-10-06 at 14 23 47

PR #212, opened 2021-11-24
Log accurate server error when SDK fails during loading process

borodovisin:
## Description
This PR improves the error notification in the console when the PayPal SDK can't be loaded. Now after we received an error message from the SDK library server (status: 400) we'll print immediately that error in the console, letting know to the user why the component can't be loaded.

Now the final users don't see any error messages when the SDK fails for a known reason. For example, if the user doesn't set a `client-id`, the request to get the PayPal SDK library fails but doesn't notify the user of the real reason. The user needs to debug the code and get the response to know about the issue with the missing `client-id`.

### Screenshots
Now the user sees this in the console: 
Screen Shot 2021-11-24 at 4 23 11 PM

And need to move to the Network tab to check what happens with the SDK request:

Screen Shot 2021-11-24 at 4 23 29 PM

With the changes the user will be able to see the problem printed in the console:
Screen Shot 2021-11-24 at 4 25 20 PM

Please leave any comment/concern/idea about these changes...

### DIFF
--- a/src/components/PayPalScriptProvider.tsx
+++ b/src/components/PayPalScriptProvider.tsx
@@ -10,15 +10,13 @@ import {
     SCRIPT_ID,
     DATA_SDK_INTEGRATION_SOURCE,
     DATA_SDK_INTEGRATION_SOURCE_VALUE,
-    ERROR_LOADING_SDK,
+    LOAD_SCRIPT_ERROR,
 } from "../constants";
 import type { ScriptProviderProps } from "../types";
 import { SCRIPT_LOADING_STATE, DISPATCH_ACTION } from "../types";
 
 /**
 This `<PayPalScriptProvider />` component handles the SDK loading process.
-The component will load the PayPal SDK library in your page transforming the internal properties into queryParams.
-That means you can set your `client-id` and the components you want to use: [buttons, messages, marks], for example.
 
 Note: You always should use this component as a wrapper for  `PayPalButtons`, `PayPalMarks`, `PayPalMessages` and `BraintreePayPalButtons` components.
  */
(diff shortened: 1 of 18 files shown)


src/constants.ts
```diff
@@ -6,6 +6,7 @@ export const DATA_CLIENT_TOKEN = "data-client-token";
 export const DATA_SDK_INTEGRATION_SOURCE = "data-sdk-integration-source";
 export const DATA_SDK_INTEGRATION_SOURCE_VALUE = "react-paypal-js";
 export const DATA_NAMESPACE = "data-namespace";
+export const LOAD_SCRIPT_ERROR = "Failed to load the PayPal JS SDK script. ";
```
gregjopa, line 9:
```suggestion
export const LOAD_SCRIPT_ERROR = "Failed to load the PayPal JS SDK script.";
```

src/components/PayPalScriptProvider.tsx
```diff
@@ -60,7 +61,8 @@ export const PayPalScriptProvider: FC<ScriptProviderProps> = ({
                     });
                 }
             })
-            .catch(() => {
+            .catch((err) => {
+                console.error(LOAD_SCRIPT_ERROR, err);
```
gregjopa, line 65:
```suggestion
                console.error(`${ LOAD_SCRIPT_ERROR } ${ err }`);
```

src/components/braintree/BraintreePayPalButtons.tsx
```diff
@@ -84,9 +61,7 @@ export const BraintreePayPalButtons: FC<BraintreePayPalButtonsComponentProps> =
                 })
                 .catch((err) => {
                     setErrorState(() => {
-                        throw new Error(
-                            `An error occurred when loading the Braintree scripts: ${err}`
-                        );
+                        throw new Error(`${LOAD_SCRIPT_ERROR}${err}`);
```
gregjopa, line 64:
```suggestion
                        throw new Error(`${ LOAD_SCRIPT_ERROR } ${ err }`);
```

Issue #214, opened 2021-12-02
[BUG]: Using the sandbox tokenized key supplied by braintree causes the BraintreePaypalButtons to fail to render


9nick9:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### 🐞 Describe the Bug

I am attempting to build the BraintreePayPalButtons but I get an error if I use the sandbox tokenized key from braintree as opposed to the client token.  Other members of my team are working on the server side to get the tokens generating so I wanted to use the static key while this was in progress. 

Code I am using:
```
const braintreeToken = 'sandbox_random_bt_string';
  const paypalToken =
    'test';

  return (
    <div>
      <PayPalScriptProvider
        options={{
          'client-id': paypalToken,
          components: 'buttons',
          'data-client-token': braintreeToken,
          intent: 'tokenize',
          vault: true,
          currency: 'USD',
        }}
      >
        <BraintreePayPalButtons
          disabled={false}
          fundingSource="paypal" // Available values are: ["paypal", "card", "credit", "paylater", "venmo"]
          createBillingAgreement={createBillingAgreement}
          onApprove={onApprove}
        />
      </PayPalScriptProvider>
    </div>
  );
```

### 😕 Current Behavior

Paypal button fails to render and I get the following error message:

Failed to render <PayPalButtons /> component. Failed to initialize:  InvalidCharacterError: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.
in PayPalButtons (created by BraintreePayPalButtons

### 🤔 Expected Behavior

Paypal button renders correctly

### 🌍 Environment

```markdown
| Software         | Version(s) |
| ---------------- | ---------- |
| react-paypal-js  |            |
| Browser          |            |
| Operating System |            |
```

### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

borodovisin:
Hi @9nick9, checking the code snippet you post above, you are facing a rendering problem because the key "data-client-token" is not a real JWT token 😢. To use the Braintree you should use a real JWT token 🔒, not a random string ✍️.

Please check out this URL: `https://paypal.github.io/react-paypal-js/?path=/docs/braintree-braintreepaypalbuttons--billing-agreement`. There we have a good example of what you are trying to accomplish. 

In the example code, we're calling this endpoint: `https://braintree-sdk-demo.herokuapp.com/api/braintree/auth` to get a new fresh JWT token and set it in the key `data-client-token`.

[...]

PR #216, opened 2021-12-03
Support Braintree tokenization keys

borodovisin:
### Description
This PR supports the data-user-id-token option key on the provider, allowing the final user to set their Braintree tokenization keys. In the previous version, the BraintreePayPalButtons component only supports data-client-token, forcing the final user to make integrations using only the JWT client tokens.

### Why are we making these changes
This change allows making the Braintree integration using either the client token or the client tokenization key. Also solve this reported bug: https://github.com/paypal/react-paypal-js/issues/214. To find more info refer to this Jira ticket: https://engineering.paypalcorp.com/jira/browse/DTPPSDK-541

### Reproduction Steps
Users can integrate the Braintree as follow:

```js
<PayPalScriptProvider
    options={{
    "client-id": "your-client-id"
     "data-client-token": "your-client-token",
     // "data-user-id-token": "tokenization-key" not supported key in current version
}}>
    <BraintreePayPalButtons {...options} />
</PayPalScriptProvider>
```
With this PR the user will be able to make the Braintree integration using a JWT client token and also using their Braintree tokenization key. A validation was set to not allow a combination of both keys. Users should choose one of those keys not both in the same provider component.

Any idea/suggestion/concern is welcome 🙏.

### DIFF
--- a/src/components/braintree/BraintreePayPalButtons.tsx
+++ b/src/components/braintree/BraintreePayPalButtons.tsx
@@ -2,7 +2,7 @@ import React, { FC, useState, useEffect } from "react";
 import { loadCustomScript } from "@paypal/paypal-js";
 
 import {
-    DATA_CLIENT_TOKEN,
+    SDK_SETTINGS,
     BRAINTREE_SOURCE,
     BRAINTREE_PAYPAL_CHECKOUT_SOURCE,
     LOAD_SCRIPT_ERROR,
@@ -38,14 +38,17 @@ export const BraintreePayPalButtons: FC<BraintreePayPalButtonsComponentProps> =
                 loadCustomScript({ url: BRAINTREE_PAYPAL_CHECKOUT_SOURCE }),
             ])
                 .then(() => {
-                    const clientToken = providerContext.options[
-                        DATA_CLIENT_TOKEN
-                    ] as string;
+                    const clientTokenizationKey: string =
+                        providerContext.options[
+                            SDK_SETTINGS.DATA_USER_ID_TOKEN
+                        ];
+                    const clientToken: string =
+                        providerContext.options[SDK_SETTINGS.DATA_CLIENT_TOKEN];
                     const braintreeNamespace = getBraintreeWindowNamespace();
 
                     return braintreeNamespace.client
                         .create({
-                            authorization: clientToken,
+                            authorization: clientTokenizationKey || clientToken,
                         })
                         .then((clientInstance) => {
                             return braintreeNamespace.paypalCheckout.create({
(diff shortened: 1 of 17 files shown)


src/constants.ts
```diff
@@ -2,17 +2,20 @@
...
+};
 export const LOAD_SCRIPT_ERROR = "Failed to load the PayPal JS SDK script.";
 
 /****************************
  * Braintree error messages *
  ****************************/
-export const EMPTY_PROVIDER_CONTEXT_CLIENT_TOKEN_ERROR_MESSAGE =
-    "A client token wasn't found in the provider parent component";
+export const EMPTY_BRAINTREE_AUTHORIZATION_ERROR_MESSAGE =
+    "Authorization data wasn't provide in the parent component, data-client-token and data-user-id-token are empty";
```
gregjopa, line 18:
```suggestion
    "Invalid authorization data. Use data-client-token or data-user-id-token to authorize.";
```

src/hooks/contextValidator.ts
```diff
@@ -31,12 +31,15 @@ export function contextNotEmptyValidator(
  * @param scriptContext the result of connecting to the context provider
  * @returns strict context avoiding null values in the type and client token
  */
-export const contextOptionClientTokenNotEmptyValidator = (
+export const optionAuthorizationNotEmpty = (
```
gregjopa, line 34: I'm wondering if we can improve on the function name. Also can you update the code comment above to mention both data attributes?
```suggestion
export const validateAuthorizationData = (
```

src/hooks/contextValidator.ts
```diff
@@ -11,7 +11,7 @@ import type { ScriptContextState } from "../types";
  * @param scriptContext the result of connecting to the context provider
  * @returns strict context avoiding null values in the type
  */
-export function contextNotEmptyValidator(
+export function optionNotEmpty(
```
gregjopa, line 14:
```suggestion
export function validateReducer(
```

src/components/braintree/BraintreePayPalButtons.tsx
```diff
@@ -38,9 +38,12 @@ export const BraintreePayPalButtons: FC<BraintreePayPalButtonsComponentProps> =
                 loadCustomScript({ url: BRAINTREE_PAYPAL_CHECKOUT_SOURCE }),
             ])
                 .then(() => {
-                    const clientToken = providerContext.options[
-                        DATA_CLIENT_TOKEN
-                    ] as string;
+                    const clientToken = (providerContext.options[
```
gregjopa, line 41: Could we break this into two statements to improve readability? Ex:
```js
const { 
    [SDK_SETTINGS.DATA_USER_ID_TOKEN]: dataUserIDToken,
    [SDK_SETTINGS.DATA_CLIENT_TOKEN]: dataClientToken,
} = providerContext.options;

const clientToken = (dataUserIDToken || dataClientToken) as string;
```

gregjopa (approved):

PR #221, opened 2021-12-20
Update README.md

mstuart:
* Added example for PayPalButtons w/ createOrder and onApprove
* "Checkout" -> "check out"


mstuart:
Tried it out on codesandbox and it works 👍

gregjopa:
Thanks @mstuart!

PR #227, opened 2022-01-06
Upgrade dependencies

borodovisin:
### Description
The PR contains the latest version for all the outdated dependencies. Except for React, now I think we'll keep version 16.

To keep sync the repo with the latest versions of the used dependencies to avoid any kind of bugs/vulnerabilities.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "homepage": "https://paypal.github.io/react-paypal-js/",
     "dependencies": {
         "@paypal/paypal-js": "^4.2.2",
-        "@paypal/sdk-constants": "^1.0.113"
+        "@paypal/sdk-constants": "^1.0.114"
     },
     "devDependencies": {
         "@babel/core": "^7.16.7",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -76,7 +76,7 @@
         "@typescript-eslint/parser": "^5.9.0",
         "babel-jest": "^27.4.6",
         "babel-loader": "^8.2.3",
-        "chalk": "^4.1.2",
+        "chalk": "^5.0.0",
```
gregjopa, line 79: Let's kee chalk at v4 since their v5 is for mjs
```suggestion
        "chalk": "^4.1.2",
```

borodovisin: yeah I saw they change to ESM.

borodovisin: It isn't the only problem. The storybook update introduces new changes and deprecates some keys on the docs.container. I need to check how to solve that issue.

gregjopa (approved):

PR #230, opened 2022-01-07
Clean eslint unnecessary comments to avoid warnings

borodovisin:
### Description
Remove es disabling comments

To avoid disabling eslint rules on our code.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,7 +28,9 @@
     },
     "rules": {
         "react/prop-types": "off",
-        "@typescript-eslint/explicit-module-boundary-types": "off"
+        "@typescript-eslint/explicit-module-boundary-types": "off",
+        "curly": "warn",
+        "eqeqeq": "warn"
     },
     "overrides": [
         {
--- a/src/components/PayPalButtons.test.tsx
+++ b/src/components/PayPalButtons.test.tsx
@@ -332,18 +332,21 @@ describe("<PayPalButtons />", () => {
         const spyConsoleError = jest
             .spyOn(console, "error")
             .mockImplementation();
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        window.paypal!.Buttons = () => {
-            return {
-                close: jest.fn().mockResolvedValue({}),
-                isEligible: jest.fn().mockReturnValue(true),
-                render: jest.fn((element: string | HTMLElement) => {
-                    // simulate adding markup for paypal button
-                    if (typeof element != "string")
-                        element.append(document.createElement("div"));
-                    return Promise.reject("Unknown error");
-                }),
-            };
+        window.paypal = {
+            Buttons() {
+                return {
+                    close: jest.fn().mockResolvedValue({}),
+                    isEligible: jest.fn().mockReturnValue(true),
+                    render: jest.fn((element: string | HTMLElement) => {
+                        // simulate adding markup for paypal button
+                        if (typeof element !== "string") {
+                            element.append(document.createElement("div"));
+                        }
+                        return Promise.reject("Unknown error");
+                    }),
+                };
+            },
+            version: "",
         };
 
         render(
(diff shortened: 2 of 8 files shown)


gregjopa (approved):

src/components/PayPalButtons.test.tsx
```diff
@@ -408,13 +410,15 @@ describe("<PayPalButtons />", () => {
...
             .fn()
             .mockRejectedValue(new Error("Unknown error"));
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        window.paypal!.Buttons = () => {
-            return {
-                close: jest.fn().mockResolvedValue({}),
-                isEligible: jest.fn().mockReturnValue(true),
-                render: mockRender,
-            };
+        window.paypal = {
```
gregjopa, line 413: Can you try this syntax instead of the arrow functions?
```suggestion
        window.paypal = {
            Buttons() {
                return {
                    close: jest.fn().mockResolvedValue({}),
                    isEligible: jest.fn().mockReturnValue(true),
                    render: mockRender,
                };
            }
        }
```

src/components/PayPalMarks.test.tsx
```diff
@@ -137,17 +137,19 @@ describe("<PayPalMarks />", () => {
...
-                    return Promise.reject("Unknown error");
-                }),
-            };
+        window.paypal = {
+            Marks() {
+                return {
+                    isEligible: jest.fn().mockReturnValue(true),
+                    render: jest.fn((element) => {
+                        // simulate adding markup for paypal mark
+                        if (typeof element != "string")
```
gregjopa, line 146: Please wrap `if` in brackets
```suggestion
                        if (typeof element != "string") {
                            element.append(document.createElement("div"));
                        }
```

src/components/PayPalButtons.test.tsx
```diff
@@ -332,18 +332,20 @@ describe("<PayPalButtons />", () => {
...
-                }),
-            };
+        window.paypal = {
+            Buttons() {
+                return {
+                    close: jest.fn().mockResolvedValue({}),
+                    isEligible: jest.fn().mockReturnValue(true),
+                    render: jest.fn((element: string | HTMLElement) => {
+                        // simulate adding markup for paypal button
+                        if (typeof element != "string")
```
gregjopa, line 342: Please use brackets and use a strict equality check
```suggestion
                        if (typeof element !== "string") {
                            element.append(document.createElement("div"));
                        }
```

src/components/PayPalMarks.test.tsx
```diff
@@ -137,17 +137,20 @@ describe("<PayPalMarks />", () => {
...
-                    return Promise.reject("Unknown error");
-                }),
-            };
+        window.paypal = {
+            Marks() {
+                return {
+                    isEligible: jest.fn().mockReturnValue(true),
+                    render: jest.fn((element) => {
+                        // simulate adding markup for paypal mark
+                        if (typeof element != "string") {
```
gregjopa, line 146:
```suggestion
                        if (typeof element !== "string") {
```

src/components/PayPalButtons.test.tsx
```diff
@@ -332,18 +332,20 @@ describe("<PayPalButtons />", () => {
...
-                        element.append(document.createElement("div"));
-                    return Promise.reject("Unknown error");
-                }),
-            };
+        window.paypal = {
+            Buttons() {
+                return {
+                    close: jest.fn().mockResolvedValue({}),
+                    isEligible: jest.fn().mockReturnValue(true),
+                    render: jest.fn((element: string | HTMLElement) => {
```
gregjopa, line 340: I just noticed that this test is meant to fail and throw a rendering error. Let's remove the conditional check and just error.
```suggestion
                    render: jest.fn((element: string | HTMLElement) => {
                       return Promise.reject("Unknown error");
                    }
```

borodovisin: We cannot remove the element coz this line on the render process: https://github.com/paypal/react-paypal-js/blob/fcaced0a17b15ec636b0b86bdf8d2a8f2a4e0983/src/components/PayPalButtons.tsx#L106

borodovisin: If the element is empty we omit the error process.

gregjopa: Ahh good point. Let's keep it then. Please use curly brackets with the `if` and strict equality with `!==`

borodovisin: Enabled rules for those cases in eslintrc.json file.

.eslintrc.json
```diff
@@ -28,7 +28,9 @@
     },
     "rules": {
         "react/prop-types": "off",
-        "@typescript-eslint/explicit-module-boundary-types": "off"
+        "@typescript-eslint/explicit-module-boundary-types": "off",
+        "curly": "warn",
```
gregjopa, line 32: @borodovisin thanks for adding these new rules! 💯

PR #231, opened 2022-01-11
Replace absolute imports types from paypal-js

borodovisin:
### Description
The PR changes the way to import types from `paypal-js`, now we can import directly from the library without need to specify the paths.

To make it cleaner the way final users import types from the `paypal-js` dependency. Now we have access to all exported types from the default `index.d.ts` in the `paypal-js` library.

### Screenshots
Previous imports:
previous

Current imports:
current

### DIFF
--- a/src/components/PayPalButtons.test.tsx
+++ b/src/components/PayPalButtons.test.tsx
@@ -13,10 +13,10 @@ import { PayPalButtons } from "./PayPalButtons";
 import { FUNDING } from "../index";
 import { loadScript, PayPalNamespace } from "@paypal/paypal-js";
 import { PayPalScriptProvider } from "./PayPalScriptProvider";
-import {
+import type {
     PayPalButtonsComponent,
     PayPalButtonsComponentOptions,
-} from "@paypal/paypal-js/types/components/buttons";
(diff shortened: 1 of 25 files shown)


src/stories/constants.ts
```diff
@@ -8,6 +8,7 @@ export const ERROR = "Error";
 export const SDK = "SDK";
 export const APPROVE = "approve";
 export const SUBSCRIPTION = "subscription";
+export const ORDER_INSTANCE_ERROR = "No order instance was provide";
```
gregjopa, line 11:
```suggestion
export const ORDER_INSTANCE_ERROR = "No order instance was provided";
```

src/components/PayPalButtons.test.tsx
```diff
@@ -16,7 +16,7 @@ import { PayPalScriptProvider } from "./PayPalScriptProvider";
 import {
     PayPalButtonsComponent,
     PayPalButtonsComponentOptions,
-} from "@paypal/paypal-js/types/components/buttons";
```
gregjopa, line 19: Should this use `import type`?

gregjopa (approved):

PR #234, opened 2022-01-13
fix: missing type billing agreement option on createPayment

jcyamacho:
the property `requestBillingAgreement` is expected in order to use the **Checkout with Vault** flow:

https://developer.paypal.com/braintree/docs/guides/paypal/checkout-with-vault


gregjopa (approved):
Thanks for the contribution @jcyamacho! 💯

Issue #215, opened 2021-12-02
[BUG]: BraintreePayPalButtons fail to render


9nick9:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### 🐞 Describe the Bug

I am trying to use the BraintreePayPalButtons as follows:

```return (
    <div>
      <PayPalScriptProvider
        options={{
          'client-id': paypalToken,
          components: 'buttons',
          'data-client-token': braintreeToken,
          intent: 'tokenize',
          vault: true,
          currency: 'USD',
        }}
      >
        <BraintreePayPalButtons
          disabled={false}
          fundingSource="paypal" // Available values are: ["paypal", "card", "credit", "paylater", "venmo"]
          createBillingAgreement={createBillingAgreement}
          onApprove={onApprove}
        />
      </PayPalScriptProvider>
    </div>
  );
```

However, this fails to render. I can get around the issue by pulling in the braintree-web npm package and pointing window.braintree to this package.  I see react-paypal-js successfully downloading the braintree scripts but something isn't initializing this on the window properly.

### 😕 Current Behavior

Paypal button fails to render with the following error message:

Failed to load the PayPal JS SDK script. TypeError: Cannot read properties of undefined (reading 'client')
in BraintreePayPalButtons (created by PaypalCheckout)

### 🤔 Expected Behavior

Paypal button renders successfully.

### 🌍 Environment

```markdown
| Software         | Version(s) |
| ---------------- | ---------- |
| react-paypal-js  |            |
| Browser          |            |
| Operating System |            |
```

### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

### ➕ Anything else?

This feels like a race condition causing the braintree script to be used before it is initialized but that is just a guess from my end.

borodovisin:
Hi @9nick9, sorry to take so long for a response. I was trying to replicate the exact same problem you're facing, unfortunately, I did not see the error message. You mention that `react-paypal-js` can successfully load the Braintree library.
That means you can see these two script loads in the Network tab: 
<img width="283" alt="Screen Shot 2021-12-14 at 3 51 05 PM" src="https://user-images.githubusercontent.com/26287838/146077380-f6fed167-4020-416f-a112-b5179e942196.png">

And when you try to access in the DevTool console to `window.braintree` it's `undefine`?
This is what you're facing now trying to use the Braintree?

borodovisin:
@9nick9 that happens all the time?. What browser are you using?. 

[...]

PR #237, opened 2022-01-17
Support loading Braintree from AMD modules

borodovisin:
### Description
The PR contains the needed modification to load the Braintree module either way, from the global `window` context or from the AMD module library. Meaning if in the global `window` scope the keys `define` and `define.amd` are defined the Braintree won't be accessible through the global `window.braintree`. Instead will be accessible via `require` module from an AMD library.

The Braintree script is supporting AMD modules, and that introduces a bug in the `react-paypal-js` repo. The main reason when this condition matches on the browser: `typeof define === "function" && define.amd` the Braintree script will be defined as an AMD module, not in the global context of the browser.
Please see this [Jira](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-630) ticket.

### Issue
This PR contains the fixes for this [issue](https://github.com/paypal/react-paypal-js/issues/215).

### DIFF
--- a/src/components/braintree/BraintreePayPalButtons.tsx
+++ b/src/components/braintree/BraintreePayPalButtons.tsx
@@ -1,17 +1,11 @@
 import React, { FC, useState, useEffect } from "react";
-import { loadCustomScript } from "@paypal/paypal-js";
 
-import {
-    SDK_SETTINGS,
-    BRAINTREE_SOURCE,
-    BRAINTREE_PAYPAL_CHECKOUT_SOURCE,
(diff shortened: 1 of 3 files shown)


gregjopa (commented):
I am concerned with adding code to this library to support AMD/require.js since its a legacy loader. 

Can this be solved in user land instead? My recommendation to folks using AMD is to write their own custom custom react component for the Braintree PayPal implementation for finer grained control over how the braintree client gets referenced.

borodovisin:
We don't need to support AMD modules on this repo.

PR #239, opened 2022-01-21
Enable deep find in PayPalHostedFieldsProvider

borodovisin:
### Description
The PR contains a solution to find all `PayPalHostedField` in the children list regardless of component depths in the `JSX` definition.

Now the `PayPalHostedFieldsProvider` component not allows to wrap `PayPalHostedField` components in deeper HTMLElements. It's hard to customize the `PayPalHostedField`, it's impossible to use the component one beside the other because the final user is not able to wrapper it inside flex div.

The PR fixes this and enables to set the `PayPalHostedField` inside any complex `JSX` structure.
More information in this [Jira](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-644) ticket.

### Reproduction Steps
The following example produces an error on the validation step:

```
<PayPalHostedFieldsProvider>
    <div>
        <PayPalHostedField />
   </div>
</PayPalHostedFieldsProvider>
```

### Screenshots
Screen Shot 2022-01-21 at 3 40 42 PM

### DIFF
--- a/src/components/hostedFields/PayPalHostedFieldsProvider.tsx
+++ b/src/components/hostedFields/PayPalHostedFieldsProvider.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef, Children } from "react";
+import React, { useState, useEffect, useRef } from "react";
 import type { FC } from "react";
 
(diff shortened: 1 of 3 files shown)


borodovisin:
This PR introduces unnecessary complex code in the logic of the Hosted Fields components.

PR #240, opened 2022-01-25
PayPalHostedField register in parent PayPalHostedFieldsProvider component

borodovisin:
### Description
The PR change the logic related to the `PayPalHostedFieldsProvider` where this component makes a recursive search in children to find the `PayPalHostedField` component and keep a track of them to use it in the HostedFields namespace render function.

I prefer to use the `useRef` hook over the `setState` hook because the `setState` will re-evaluate all the time the `PayPalHostedFieldsProvider` component each time children register and call to the `set` function. I don't like to re-evaluate all the time when a child component is registered in the parent and put an escape if until we have all the children registered. 

Based on the idea React is a declarative(code) library, the execution of the `useEffect` will follow the same behavior all the time, executing the `useEffect` in the deepest children before the `useEffect` in the parent component. Meaning it's safe to register the `PayPalHostedField` in the mounted lifecycle, avoiding any kind of race condition in the `useEffect` executed in the `PayPalHostedFieldsProvider` component, and avoiding unnecessary re-evaluations.

This change substitutes the previous logic where the parent component has the responsibility to search and find all the `PayPalHostedField` components. Making the code hard to maintain and understand due to the recursive function to search in the children tree. 

The new logic changes the responsibility and now each child component has the responsibility to register in the parent component.

More details in this [JIRA](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-645) ticket.

### Breaking changes
The Pr introduce a braking change on the `usePayPalHostedFields` custom hook. In the previous version, the hook returns directly the `cardFields` instance. Now to get access to the `cardFields` the final user should destructure the resulting object. 
Prevoiously: `const cardFields = usePayPalHostedFields()`;
Current: `const { cardFields } = usePayPalHostedFields()`;


src/components/hostedFields/utils.ts
```diff
@@ -46,38 +35,66 @@ export const generateMissingHostedFieldsError = ({
 };
 
 /**
- * Identify all the valid hosted fields children and generate the valid options
- * to use in the HostedFields.render process
+ * Validate the expiration date children. Valid combinations are:
```
gregjopa, line 38:
```suggestion
 * Validate the expiration date component. Valid combinations are:
```

src/components/hostedFields/utils.ts
```diff
@@ -46,38 +35,66 @@ export const generateMissingHostedFieldsError = ({
...
+ * 1- Exists expirationDate field and not exists expirationMonth and expirationYear fields
```
gregjopa, line 39:
```suggestion
 * 1 - only the `expirationDate` field exists. 
```

src/components/hostedFields/utils.ts
```diff
@@ -46,38 +35,66 @@ export const generateMissingHostedFieldsError = ({
...
+ * 2- Exists expirationMonth and expirationYear and not exists expirationDate
```
gregjopa, line 40:
```suggestion
 * 2 - only the `expirationMonth` and `expirationYear` fields exist. Cannot be used with the `expirationDate` field.
```

src/components/hostedFields/utils.ts
```diff
@@ -46,38 +35,66 @@ export const generateMissingHostedFieldsError = ({
...
+ * All the other possible combinations are invalid
```
gregjopa, line 41:
```suggestion
```

src/components/hostedFields/utils.ts
```diff
@@ -46,38 +35,66 @@ export const generateMissingHostedFieldsError = ({
...
+ * @throw an @type {Error} when duplicate types was found
+ */
+const notDuplicateChildren = (registerTypes: PAYPAL_HOSTED_FIELDS_TYPES[]) => {
+    if (registerTypes.length !== new Set(registerTypes).size) {
+        throw new Error(HOSTED_FIELDS_DUPLICATE_CHILDREN_ERROR);
+    }
+};
+
+/**
+ * Validate the hosted fields children past to the PayPalHostedFieldsProvider component
```
gregjopa, line 86:
```suggestion
 * Validate the hosted field children in the PayPalHostedFieldsProvider component.
```

src/components/hostedFields/utils.ts
```diff
@@ -46,38 +35,66 @@ export const generateMissingHostedFieldsError = ({
...
+ * These are the rule:
```
gregjopa, line 87:
```suggestion
 * These are the rules:
```

src/components/hostedFields/utils.ts
```diff
@@ -46,38 +35,66 @@ export const generateMissingHostedFieldsError = ({
...
+ * 1- We need to find 3 default children type [[number, expiration, cvv]
```
gregjopa, line 88:
```suggestion
 * 1 - We need to find 3 default children for number, expiration, cvv
```

src/components/hostedFields/utils.ts
```diff
@@ -46,38 +35,66 @@ export const generateMissingHostedFieldsError = ({
...
+ * 2- We cannot find duplicate children type
```
gregjopa, line 89:
```suggestion
 * 2 - No duplicate children are allowed.
```

src/components/hostedFields/utils.ts
```diff
@@ -46,38 +35,66 @@ export const generateMissingHostedFieldsError = ({
...
+ * 3- We cannot find expirationDate field combine with expirationMonth and expirationYear
```
gregjopa, line 90:
```suggestion
 * 3 - No invalid combinations of `expirationDate`, `expirationMonth`, and `expirationYear`
```

src/components/hostedFields/utils.ts
```diff
@@ -46,38 +35,66 @@ export const generateMissingHostedFieldsError = ({
...
+ *
+ * @param childrenList     the list of children
+ * @param requiredChildren the list with required children [number, expiration, cvv]
+ */
+export const validateHostedFieldChildren = (
+    registeredFields: PAYPAL_HOSTED_FIELDS_TYPES[]
+): void => {
+    hasDefaultChildren(registeredFields);
+    notDuplicateChildren(registeredFields);
```
gregjopa, line 99:
```suggestion
    noDuplicateChildren(registeredFields);
```

src/components/hostedFields/utils.ts
```diff
@@ -46,38 +35,66 @@ export const generateMissingHostedFieldsError = ({
...
-        }
-        return fields;
-    }, {});
+/**
+ * Check if we don't have duplicate children types
+ *
+ * @param registerTypes the list of all the children types pass to the parent
+ * @throw an @type {Error} when duplicate types was found
+ */
+const notDuplicateChildren = (registerTypes: PAYPAL_HOSTED_FIELDS_TYPES[]) => {
```
gregjopa, line 79:
```suggestion
const noDuplicateChildren = (registerTypes: PAYPAL_HOSTED_FIELDS_TYPES[]) => {
```

src/components/hostedFields/PayPalHostedFieldsProvider.tsx
```diff
@@ -77,10 +78,13 @@ export const PayPalHostedFieldsProvider: FC<
                 // Call your server to set up the transaction
                 createOrder: createOrder,
                 styles: styles,
-                fields: generateHostedFieldsFromChildren(childrenList),
+                fields: registeredFields.current,
             })
             .then((cardFieldsInstance) => {
-                setCardFields(cardFieldsInstance);
+                // Avoid memory leak trying to set value on unmount component
```
gregjopa, line 84:
```suggestion
```

gregjopa: I think its okay to remove this comment. We do a similar thing in Buttons component where we ensure the ref is valid before using it.

src/components/hostedFields/hooks.ts
```diff
@@ -0,0 +1,30 @@
+import { useRef, MutableRefObject } from "react";
+
+import type { PayPalHostedFieldsRegistered } from "../../types/payPalHostedFieldTypes";
+
+/**
+ * Custom hook to store registered hosted fields children
+ * Each `PayPalHostedField` component should be registered on the parent provider
+ *
+ * @since 7.5.1
```
gregjopa, line 9: I'd like to avoid using the `@since` in the java doc style sections. The 7.5.1 version is technically wrong. We don't know the next version this will be included in. It could be `7.5.2` or `7.6.0` depending on the semver changes.
```suggestion
```

src/components/hostedFields/utils.test.js
```diff
@@ -41,50 +40,51 @@ describe("generateMissingHostedFieldsError", () => {
...
 
-    test("should return empty object when children argument doesn't have PayPalHostedField components", () => {
-        expect(
-            generateHostedFieldsFromChildren([
-                <div key="0"></div>,
-                <input key="1" />,
-                <button key="2">Submit</button>,
-            ])
-        ).toEqual({});
+    test("should fail when cvv and card number is not a child", () => {
```
gregjopa, line 50:
```suggestion
    test("should fail when missing children for cvv and card number", () => {
```

src/components/hostedFields/utils.test.js
```diff
@@ -41,50 +40,51 @@ describe("generateMissingHostedFieldsError", () => {
...
-                selector: ".car-number",
-            },
-            expirationDate: {
-                selector: ".expiration",
-            },
-            cvv: {
-                selector: ".cvv",
-            },
-        });
+    test("should fail when expiration is not a child", () => {
```
gregjopa, line 59:
```suggestion
    test("should fail when missing children for expiration", () => {
```

gregjopa (approved):

PR #242, opened 2022-01-27
Avoid null comparisons in the code

borodovisin:
### Description
The PR avoids the null comparison in the code of the repo.

To make the code more readable and easier to understand.


gregjopa:
I would prefer to keep these explicit null checks since we default these useRef/useState values as `null`

borodovisin:
> useRef/useState values as `null`

Sounds good to me...

PR #252, opened 2022-02-08
chore: HostedFields stories are failing

borodovisin:
### Description
The PR fix a problem with the hosted fields stories. The `createOrder` function is failing since is sending an `undefined` value in `currency_code`.

To fix the broken hosted fields stories.


wsbrunson (approved):
good looking out!

gregjopa (approved):
Nice catch @borodovisin!

PR #265, opened 2022-05-06
Support merchant-id on checkout creation

borodovisin:
### Description
The PR contains an enhancement on the Braintree integration. A new missing case was added to the React Braintree integration allowing the use of the `merchant-id` in the `paypalCheckout.create` function initialization.

In the current version, the solution only supports the client instance, but the final clients aren't able to use the `merchantAccountId` option on the checkout initialization process.
More details [here](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-754)

### DIFF
--- a/src/components/braintree/utils.ts
+++ b/src/components/braintree/utils.ts
@@ -4,6 +4,7 @@ import { getBraintreeWindowNamespace } from "../../utils";
 import {
     BRAINTREE_SOURCE,
     BRAINTREE_PAYPAL_CHECKOUT_SOURCE,
+    BRAINTREE_MULTIPLE_MERCHANT_IDS_ERROR_MESSAGE,
 } from "../../constants";
 
 import type { BraintreeNamespace } from "./../../types/braintreePayPalButtonTypes";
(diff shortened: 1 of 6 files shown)


src/types/braintree/paypalCheckout.ts
```diff
@@ -89,9 +89,9 @@ export interface BraintreePayPalCheckout {
      * });
      */
     create(options: {
-        client?: BraintreeClient | undefined;
-        authorization?: string | undefined;
-        merchantAccountId?: string | undefined;
+        client?: BraintreeClient;
+        authorization?: string;
+        merchantAccountId?: Array<string> | string;
```
gregjopa, line 94: It seems like braintree-web only supports a single merchant id so I would go with this typing.
```suggestion
        merchantAccountId?: string;
```

src/constants.ts
```diff
@@ -16,6 +17,8 @@ export const LOAD_SCRIPT_ERROR = "Failed to load the PayPal JS SDK script.";
  ****************************/
 export const EMPTY_BRAINTREE_AUTHORIZATION_ERROR_MESSAGE =
     "Invalid authorization data. Use data-client-token or data-user-id-token to authorize.";
+export const BRAINTREE_MULTIPLE_MERCHANT_IDS =
+    "Braintree PayPal checkout can be configured with only one merchantId, instead multiple were received";
```
gregjopa, line 21:
```suggestion
    "Braintree does not support passing in multiple merchantId values.
```

src/components/braintree/utils.ts
```diff
@@ -97,3 +98,22 @@ export const getBraintreeNamespace = (
...
+export const getMerchantId = (
+    source?: Array<string> | string
+): string | undefined => {
+    const isSourceArray = Array.isArray(source);
+
+    if (isSourceArray && source.length > 1) {
+        throw new Error(BRAINTREE_MULTIPLE_MERCHANT_IDS);
+    }
+
+    return isSourceArray ? (source || [])[0] : source;
```
gregjopa, line 118: Can you use toString() to take care of converting an array of size 1 or a string to a string?
```suggestion
    return source.toString();
```

src/constants.ts
```diff
@@ -16,6 +17,8 @@ export const LOAD_SCRIPT_ERROR = "Failed to load the PayPal JS SDK script.";
  ****************************/
 export const EMPTY_BRAINTREE_AUTHORIZATION_ERROR_MESSAGE =
     "Invalid authorization data. Use data-client-token or data-user-id-token to authorize.";
+export const BRAINTREE_MULTIPLE_MERCHANT_IDS =
```
gregjopa, line 20:
```suggestion
export const BRAINTREE_MULTIPLE_MERCHANT_IDS_ERROR_MESSAGE =
```

wsbrunson (approved):

PR #268, opened 2022-05-09
chore: minor edit on Braintree merchant-id

borodovisin:
### Description
The PR contains a fix to avoid sending empty strings in the merchant-id options.

To avoid issues creating a payment with an empty merchant-id string.


borodovisin:
This PR is not needed anymore...

PR #272, opened 2022-05-23
chore(story): update the Hosted Fields story example

borodovisin:
### Description
The PR includes a code change about how to use the Hosted Fields. Instead of using an alias changing the `cardFields` name to `hostedField`, the idea is to go straight with the instance and the key as follows: `hostedField.cardFields`.

The example code for Hosted Fields can introduce some confusion because we're using an alias on line 75 of the file `PayPalHostedFields.stories.tsx` take a look [here](https://github.com/paypal/react-paypal-js/blob/904c580046c1f7d3d83311ab735f1a56c856f8ff/src/stories/payPalHostedFields/PayPalHostedFields.stories.tsx#L75)
More details [here](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-778)

### Screenshots
Before:
before

After:
after

### DIFF
--- a/src/stories/payPalHostedFields/PayPalHostedFields.stories.tsx
+++ b/src/stories/payPalHostedFields/PayPalHostedFields.stories.tsx
@@ -72,51 +72,54 @@ function captureOrderUrl(orderId: string): string {
 const SubmitPayment = ({ customStyle }: { customStyle?: CSSProperties }) => {
     const [paying, setPaying] = useState(false);
     const cardHolderName = useRef<HTMLInputElement>(null);
-    const { cardFields: hostedField } = usePayPalHostedFields();
+    const hostedField = usePayPalHostedFields();
 
     const handleClick = () => {
-        if (hostedField) {
-            if (
-                Object.values(hostedField.getState().fields).some(
-                    (field) => !field.isValid
-                ) ||
-                !cardHolderName?.current?.value
-            ) {
-                return alert(
-                    "The payment form is invalid, please check it before execute the payment"
-                );
-            }
-            action(SUBMIT_FORM)("Form is valid and submitted");
-            setPaying(true);
-            hostedField
-                .submit({
-                    cardholderName: cardHolderName?.current?.value,
-                })
-                .then((data) => {
-                    action(`Received ${ORDER_ID}`)(data.orderId);
-                    action(CAPTURE_ORDER)(
-                        `Sending ${ORDER_ID} to custom endpoint to capture the payment information`
-                    );
-                    fetch(captureOrderUrl(data.orderId), {
-                        method: "post",
-                    })
-                        .then((response) => response.json())
-                        .then((data) => {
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
+### Description
```
gregjopa, line 1: I like this PR template, but can we add this in a separate PR?

borodovisin: yeah, I'll get rid of this file for this PR

src/stories/payPalHostedFields/PayPalHostedFields.stories.tsx
```diff
@@ -72,12 +72,12 @@ function captureOrderUrl(orderId: string): string {
 const SubmitPayment = ({ customStyle }: { customStyle?: CSSProperties }) => {
     const [paying, setPaying] = useState(false);
     const cardHolderName = useRef<HTMLInputElement>(null);
-    const { cardFields: hostedField } = usePayPalHostedFields();
+    const hostedField = usePayPalHostedFields();
 
     const handleClick = () => {
-        if (hostedField) {
+        if (hostedField?.cardFields) {
```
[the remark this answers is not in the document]
borodovisin: Make sense, I'll make the change

src/stories/payPalHostedFields/PayPalHostedFields.stories.tsx
```diff
@@ -72,51 +72,57 @@ function captureOrderUrl(orderId: string): string {
...
+        if (!hostedField?.cardFields) {
+            const childErrorMessage =
+                "Unable to find any child components in the <PayPalHostedFieldsProvider />";
+
+            action(ERROR)(childErrorMessage);
+            throw new Error(childErrorMessage);
+        }
+
+        if (
+            Object.values(hostedField.cardFields.getState().fields).some(
```
gregjopa, line 87: Could we move this one to a variable to simplify the `if` condition?

```js
const hasInvalidField = Object.values(hostedField.cardFields.getState().fields).some((field) => !field.isValid);

if (hasInvalidField || !cardHolderName?.current?.value) {
    return alert("Invalid payment form");
}

```

src/stories/payPalHostedFields/PayPalHostedFields.stories.tsx
```diff
@@ -72,51 +72,56 @@ function captureOrderUrl(orderId: string): string {
...
+            throw new Error(childErrorMessage);
+        }
+        const isFormInvalid =
+            Object.values(hostedField.cardFields.getState().fields).some(
+                (field) => !field.isValid
+            ) || !cardHolderName?.current?.value;
+
+        if (isFormInvalid) {
+            return alert(
+                "The payment form is invalid, please check it before execute the payment"
```
gregjopa, line 92: I would prefer to simplify this error message:
```suggestion
                "The payment form is invalid"
```

PR #277, opened 2022-05-27
chore(typescript): remove type from import

borodovisin:
### Description
Remove the `import { type TYPE }` to avoid problems with an older version of Typescript.

Check details [here](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-782)

### DIFF
--- a/src/components/PayPalButtons.test.tsx
+++ b/src/components/PayPalButtons.test.tsx
@@ -1,4 +1,4 @@
-import React, { useState, type ReactNode } from "react";
+import React, { useState } from "react";
 import {
     render,
     waitFor,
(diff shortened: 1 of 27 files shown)


wsbrunson (approved):

src/components/PayPalButtons.test.tsx
```diff
@@ -8,17 +8,18 @@ import {
 } from "@testing-library/react";
 import { ErrorBoundary } from "react-error-boundary";
 import { mock } from "jest-mock-extended";
-import {
-    loadScript,
-    PayPalNamespace,
-    type PayPalButtonsComponent,
-    type PayPalButtonsComponentOptions,
-} from "@paypal/paypal-js";
+import { loadScript, PayPalNamespace } from "@paypal/paypal-js";
```
gregjopa, line 11: Technically `PayPalNamespace` is a type as well

PR #310, opened 2022-10-02
fix: add memoization in PayPalScriptProvider to break infinite loop

### LINKED ISSUE #287: [BUG]: Continuous call of host page
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### 🐞 Describe the Bug

Having implemented this in React TS - I then embed in asp.net core 6 page.

### 😕 Current Behavior

This causes the page to endlessly loop into a reload because this component is calling a create method and requesting the information from paypal - that method then creates a browser redirect message which intern gets asp.net to call the action method of the controller that builds the html page - therefore loading again and starting your component again. This is unusable.

### 🤔 Expected Behavior

On load - all elements should be loaded along with the plugin without validation through papal.

### 🔬 Minimal Reproduction

_No response_

### 🌍 Environment

```markdown
| Software         | Version(s) |
| ---------------- | ---------- |
| react-paypal-js  |            |
| Browser          |            |
| Operating System |            |
```

### Relevant log output

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### ➕ Anything else?

_No response_

pfedorowiat:
ISSUE: 
Using the `dispatch` function from the `usePayPalScriptReducer` hook in the `useEffect` callback is causing the infinity loop. 

Infinity loop is causing a massive amount of resource fetch which is resulting in throttling (429 HTTP code).

EXAMPLE: 
Code example available on https://paypal.github.io/react-paypal-js/?path=/docs/example-paypalbuttons--default does not trigger this issue because the `useEffect` dependency array is incomplete. 
``` 
    const [{ options, isPending }, dispatch] = usePayPalScriptReducer();

    useEffect(() => {
        dispatch({
            type: "resetOptions",
            value: {
                ...options,
                currency: currency,
            },
        });
    }, [currency, showSpinner]); // <-- missing `options` dependency
```
Skipping items of dependency array is not a good practice, as explained here https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528

Fixing that will trigger the loop.
``` 
    const [{ options, isPending }, dispatch] = usePayPalScriptReducer();

    useEffect(() => {
        dispatch({
            type: "resetOptions",
            value: {
                ...options,
                currency: currency,
            },
        });
    }, [currency, options, showSpinner]); 
```

SOLUTION:
This PR is adding memoization of the `options` object. The comparison between the current and previous values of the `options` object is made to ensure that the object will change its reference only when its value changes.

Replacing the `useReducer` hook with multiple `useState` hooks for each state element and exposing callback functions instead of `dispatch` will result in better isolation of state changes. Also, it should prevent similar issues in the future. However, it will be a breaking change, so this PR is limited to fix of only one issue.

OTHER:
Probably fixes #287

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@paypal/react-paypal-js",
-    "version": "7.8.1",
+    "version": "7.8.2",
     "description": "React components for the PayPal JS SDK",
     "keywords": [
         "react",
--- a/src/components/PayPalScriptProvider.tsx
+++ b/src/components/PayPalScriptProvider.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useReducer } from "react";
+import React, { useEffect, useMemo, useReducer, useRef } from "react";
 import { loadScript } from "@paypal/paypal-js";
 
 import {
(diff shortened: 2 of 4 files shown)


gregjopa:
Hi @pfedorowiat, can you provide a link to a demo react app that shows the infinite loop problem? Let’s start with a reproduction of the problem first before working on a solution.

PR #313, opened 2022-11-29
Update main deps

gregjopa:


src/index.ts
```diff
@@ -10,10 +10,14 @@ export * from "./components/PayPalScriptProvider";
...
+import type { FUNDING_SOURCE } from "@paypal/paypal-js";
 import * as constants from "@paypal/sdk-constants/dist/module";
 
 // We do not re-export `FUNDING` from the `sdk-constants` module
 // directly because it has no type definitions.
 //
 // See https://github.com/paypal/react-paypal-js/issues/125
-export const FUNDING: Record<string, string> = constants.FUNDING;
+export const FUNDING = constants.FUNDING as Record<
+    Uppercase<FUNDING_SOURCE>,
```
gregjopa, line 21: Previously our FUNDING constants didn't have types and were just strings. Now they should be fully typed out.

prashantjagasia (approved):
Looks good to me 👍

westeezy (approved):

package.json
```diff
@@ -45,8 +45,8 @@
     },
     "homepage": "https://paypal.github.io/react-paypal-js/",
     "dependencies": {
-        "@paypal/paypal-js": "^5.0.6",
-        "@paypal/sdk-constants": "^1.0.116"
+        "@paypal/paypal-js": "^5.1.4",
+        "@paypal/sdk-constants": "^1.0.122"
```
westeezy, line 49: We noticed a few libs that had outdated sdk-constants before. I wonder if we can get renovate or some such wired in to keep up with that better. It doesn't seem like a major issue, but could be a nice way to remind us to update any packages we miss. We could set it to auto merge on sdk-constants for instance if tests pass

wsbrunson (approved):

PR #366, opened 2023-06-29
feat: add loadingStatusErrorMessage to state

dtjones404:
Probably naive first attempt to expose error message in case of rejected script load

### DIFF
--- a/src/types/enums.ts
+++ b/src/types/enums.ts
@@ -17,7 +17,6 @@ export enum SCRIPT_LOADING_STATE {
  */
 export enum DISPATCH_ACTION {
     LOADING_STATUS = "setLoadingStatus",
-    LOADING_STATUS_ERROR_MESSAGE = "loadingStatusErrorMessage",
     RESET_OPTIONS = "resetOptions",
     SET_BRAINTREE_INSTANCE = "braintreeInstance",
 }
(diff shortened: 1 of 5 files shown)


src/types/enums.ts
```diff
@@ -17,6 +17,7 @@ export enum SCRIPT_LOADING_STATE {
  */
 export enum DISPATCH_ACTION {
     LOADING_STATUS = "setLoadingStatus",
+    CRITICAL_ERROR = "criticalError",
```
gregjopa, line 20: What do you think about this name? It may be nice to name it after LOADING_STATUS so it's clearer that this error message is related to a script loading failure.
```suggestion
    LOADING_STATUS_ERROR_MESSAGE = "loadingStatusErrorMessage",
```

dtjones404: Good call!  Renamed

src/types/scriptProviderTypes.ts
```diff
@@ -13,6 +13,14 @@ export type ScriptReducerAction =
           type: `${DISPATCH_ACTION.LOADING_STATUS}`;
           value: SCRIPT_LOADING_STATE;
       }
+    | {
+          type: `${DISPATCH_ACTION.LOADING_STATUS}`;
+          value: { state: SCRIPT_LOADING_STATE; message: string };
+      }
+    | {
+          type: `${DISPATCH_ACTION.LOADING_STATUS_ERROR_MESSAGE}`;
```
gregjopa, line 21: Do we still need this `type`? Can we remove it?

gregjopa:
Great work @dtjones404! 💯

PR #378, opened 2023-08-16
fix: allow users to change data-sdk-integration-source

dtjones404:
* reordered object spreading to allow users to change `data-sdk-integration-source` parameter


wsbrunson (approved):

gregjopa (approved):
💯

PR #428, opened 2024-04-11
Feature/card fields v2 support

sebastianfdz:
### This PR aims to add support for `CardFields` V2 to our react adapter.

This is a pretty dense PR. I am making a single PR because the Context Provider and the Individual components are intertwined and I think this gives all the context to be able to review it.

[This document includes some context on these changes](https://paypal.atlassian.net/wiki/spaces/MI/pages/811303598/React+HCF+V2+Support)

**### TL'DR**
There are 5 new major components added to the library:
1. **PayPalCardFieldsProvider**
Context provider the instantiates `window.paypal.CardFields` and keeps track of the state of each individual CardField 
2. **PayPalCVVField**
Individual component in charge of instantiating and rendering the CVVField. Eg: `cardFields.CVVField().render()`
3. **PayPalExpiryField**
Individual component in charge of instantiating and rendering the ExpiryField. Eg: `cardFields.ExpiryField().render()`
4. **PayPalNameField**
Individual component in charge of instantiating and rendering the NameField. Eg: `cardFields.NameField().render()`
5. **PayPalNumberField**
Individual component in charge of instantiating and rendering the NumberField. Eg: `cardFields.NumberField().render()`

The `README.md` includes the example merchant integration code for the above.

Screenshot of the integration code in the README.md when ran locally.
Screenshot 2024-04-11 at 12 32 09 p m

Tickets:
- https://paypal.atlassian.net/browse/DTPPCPSDK-2091
- https://paypal.atlassian.net/browse/DTPPCPSDK-2092
- https://paypal.atlassian.net/browse/DTPPCPSDK-2096
- https://paypal.atlassian.net/browse/DTPPCPSDK-2097


README.md
```diff
@@ -360,6 +360,84 @@ export default function App() {
 }
 ```
 
+### PayPal Card Fields
```
wsbrunson, line 363: excellent

src/components/cardFields/PayPalCVVField.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    useEffect(() => {
+        if (!cardFields) {
+            setError(() => {
+                throw new Error(
+                    "Individual CardFields must be rendered inside the PayPalCardFieldsProvider"
+                );
+            });
+            return closeComponent;
+        }
```
wsbrunson, line 31: I would consider creating a hook for these lines and share it with all the children fields

src/components/cardFields/PayPalCardFieldsProvider.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const { registerField, unregisterField } = usePayPalCardFieldsRegistry();
+
+    const [cardFields, setCardFields] =
+        useState<PayPalCardFieldsComponent | null>(null);
+    const cardFieldsInstance = useRef<PayPalCardFieldsComponent | null>(null);
+
+    const nameField = useRef<PayPalCardFieldsIndividualField | null>(null);
+    const numberField = useRef<PayPalCardFieldsIndividualField | null>(null);
+    const cvvField = useRef<PayPalCardFieldsIndividualField | null>(null);
+    const expiryField = useRef<PayPalCardFieldsIndividualField | null>(null);
```
wsbrunson, line 35: since we are using a registry (great pattern!), do you think we can find a way that the `Provider` of the card fields component doesn't need to know about all the children? 

we let the children register themselves so if we ever add or remove a component we wouldn't need to update these lines here

wsbrunson (commented):
All of the child card field components have very similar implementations. The only thing that changes between them is the name. I wonder if there's a way to create a base card child component and have each of them render that with the right names & refs

src/components/cardFields/PayPalCVVField.tsx
sebastianfdz: That would definitely be a good abstraction! 👍

README.md
```diff
@@ -360,6 +360,84 @@ export default function App() {
...
+        // merchant code
+    }
+    function onError() {
+        // merchant code
+    }
+    return (
+        <PayPalScriptProvider
+            options={{
+                clientId: "your-client-id",
+                dataClientToken: "your-data-client-token",
```
gregjopa, line 420: Do we need dataClientToken for card-fields? I thought only v1 used it.
```suggestion
```

src/components/cardFields/PayPalCardFieldsProvider.tsx
sebastianfdz: I think that would definitely make it more maintainable. One thing I'm thinking about is that we are giving merchants access to our `usePayPalCardFields` hook which is our `useContext` under the hood. Through this hook we give them access the each field's ref so that they can interact with them:
Eg: `const { cvvField } = usePayPayCardFields()`
I think we couldn't expose those refs if the provider wasn't aware of them?

README.md
sebastianfdz: Was it removed in favor of `dataUserIdToken`? I'm running that provider locally and it seems to be expected one of the two 🤔.

gregjopa: Yes I think data-client-token is only for hosted-fields (v1) and we can remove it.

For card-fields (v2). The one time checkout flow doesn't need any token passed as a data attribute. Here's an example: https://github.com/paypal-examples/docs-examples/blob/main/advanced-integration/v2/server/views/checkout.ejs#L20.

I'm not sure if card fields needs the dataUserIdToken for the vaulted flows. I don't see it mentioned in the docs (https://developer.paypal.com/docs/checkout/save-payment-methods/during-purchase/js-sdk/cards/). @wsbrunson do you happen to know?

sebastianfdz:
> All of the child card field components have very similar implementations. The only thing that changes between them is the name. I wonder if there's a way to create a base card child component and have each of them render that with the right names & refs

That would be a great abstraction! One thing that comes to mind is if we accept the same props for all of the fields? I am aware of a `maxLength` for NameField for example. If we don't anticipate having many different options for each field, I think it would absolutely make sense to have some shared logic. What are your thoughts @wsbrunson ? Also, thinking about it again, we would transparently pass the props to the shared component and if it doesn't exist (like `maxLength` in cvv) then it would just get ignored?

README.md
wsbrunson: @elizabethmv I think we are recommending partners pass this (CP4 partners specifically). Since these are new though, I would use our new `sdkClientToken`

wsbrunson: But right now CardFields does not have a return buyer experience so there's no need for any tokens for regular merchants

elizabethmv: `dataClientToken` should be removed. For `dataUserIdToken`- thats correct right now cp4 partners would pass it for the vault with/without purchase scenarios. Maybe we can leave it off for now and add the `sdkClientToken` in a follow up

src/components/cardFields/PayPalCardFieldsProvider.tsx
sebastianfdz: And by interact I mean something like:
`const { cvvField } = usePayPayCardFields()`
`cvv.current.focus()`

wsbrunson: what if in `context` there was `fields: {}`. it would be an empty object at first and as fields register, they will register inside that object.

I don't think that Provider is using the field refs other than passing them in there but I may be missing it

sebastianfdz: Dang! I think that could work. I like the concept of each field "self-registering" with its own field instance 😃. I'll do some tests.

README.md
sebastianfdz: Got it!

src/components/cardFields/PayPalCardFieldsProvider.tsx
sebastianfdz: Thanks for the great feedback @wsbrunson ! I moved the state management of registered fields to the `usePayPalCardFieldsRegistry` hook. This will allow the provider to just be aware of `fields` and pass them on as they are registered.

src/components/cardFields/PayPalCVVField.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+// import { usePayPalCardFields } from "./hooks";
+import { PayPalCardFieldsIndividualFieldOptions } from "../../types";
+import { PayPalCardField } from "./PayPalCardField";
+
+export const PayPalCVVField: React.FC<
+    PayPalCardFieldsIndividualFieldOptions
+> = (options) => {
+    // const { cvvField } = usePayPalCardFields();
```
wsbrunson, line 10: can we delete this?

src/components/cardFields/PayPalCVVField.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+    return (
+        <PayPalCardField
+            // fieldRef={cvvField}
```
wsbrunson, line 14: oh and this

src/components/cardFields/PayPalExpiryField.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+// import { usePayPalCardFields } from "./hooks";
+import { PayPalCardFieldsIndividualFieldOptions } from "../../types";
+import { PayPalCardField } from "./PayPalCardField";
+
+export const PayPalExpiryField: React.FC<
+    PayPalCardFieldsIndividualFieldOptions
+> = (options) => {
+    // const { expiryField } = usePayPalCardFields();
```
wsbrunson, line 10: same with these. i think they are in all the fields components

src/components/cardFields/context.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export type PayPalCardFieldsContextType = {
+    cardFieldsForm: PayPalCardFieldsComponent | null;
+} & RegistryHookReturnType;
+
+export const PayPalCardFieldsContext =
+    createContext<PayPalCardFieldsContextType>({
+        cardFieldsForm: null,
+        fields: {},
+        registerField: ignore, // implementation is inside hook and passed through the provider
```
wsbrunson, line 16: TIL about `ignore`

wsbrunson (approved):
super super minor feedback, this looks great!

src/components/cardFields/__snapshots__/PayPalCardFieldsProvider.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`PayPalCardFieldsProvider should throw an error when the 'card-fields' component is missing from the components list passed to the PayPalScriptProvider 1`] = `"Invalid authorization data. Use dataClientToken or dataUserIdToken to authorize."`;
```
elizabethmv, line 3: `Invalid authorization data. Use dataClientToken or dataUserIdToken to authorize`
Can we double check this? Neither dataClientToken or dataUserIdToken should be required to authorize

README.md
```diff
@@ -360,6 +360,84 @@ export default function App() {
...
+        // merchant code
+    }
+    function onError() {
+        // merchant code
+    }
+    return (
+        <PayPalScriptProvider
+            options={{
+                clientId: "your-client-id",
+                dataUserIdToken: "your-data-user-id-token",
```
elizabethmv, line 420:
```suggestion
```

I think this snuck back in

src/components/cardFields/PayPalCVVField.tsx
sebastianfdz: Ah yes! Forgot to delete those 👍

src/components/cardFields/__snapshots__/PayPalCardFieldsProvider.test.tsx.snap
sebastianfdz: This validation comes from here https://github.com/paypal/react-paypal-js/blob/main/src/hooks/contextValidator.ts#L42 . Should this not be required only for CardFields? Or in general?

sebastianfdz: It is currently used by hosted-fields v1. But Buttons uses a different hook (which doesn't make the token validation), I'm thinking we should be using the same as Buttons?

sebastianfdz: Ok, removed the token validation used in hosted-fields v1! Thanks for pointing that out again for me @elizabethmv 😃

PR #429, opened 2024-04-18
Feature/add card fields form

sebastianfdz:
Ticket: DTPPCPSDK-2102

This PR intends to add our `PayPalCardFieldsForm` for merchants to use as a standalone component. Merchants will still have access to our `usePayPalCardFields` hook to handle their own submit logic and render their own submit button with their own styling.

Screenshot 2024-04-18 at 11 00 16 a m

### DIFF
--- a/src/components/cardFields/PayPalCardFieldsForm.tsx
+++ b/src/components/cardFields/PayPalCardFieldsForm.tsx
@@ -1,0 +1,36 @@
+import React from "react";
+
+import { PayPalCardFieldsFormOptions } from "../../types";
+import { PayPalCardField } from "./PayPalCardField";
+import { FlexContainer } from "../ui/FlexContainer";
+
+export const PayPalCardFieldsForm: React.FC<PayPalCardFieldsFormOptions> = ({
+    className,
+    ...options
+}) => {
+    return (
+        <>
+            <style
+                dangerouslySetInnerHTML={{
+                    __html: `.w-full { width:100%; }`,
+                }}
+            />
+            <div className={className}>
+                <PayPalCardField fieldName="NameField" {...options} />
+                <PayPalCardField fieldName="NumberField" {...options} />
+                <FlexContainer>
+                    <PayPalCardField
+                        fieldName="ExpiryField"
+                        className="w-full"
+                        {...options}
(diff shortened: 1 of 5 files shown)


src/components/cardFields/PayPalCardFieldsForm.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { FlexContainer } from "../ui/FlexContainer";
+
+export const PayPalCardFieldsForm: React.FC<PayPalCardFieldsFormOptions> = ({
+    className,
+    ...options
+}) => {
+    return (
+        <>
+            <style
+                dangerouslySetInnerHTML={{
```
wsbrunson, line 14: 😰 😰 😰

wsbrunson: I guess we don't have any css in js for this library

wsbrunson (approved):
I think this all looks good, we should chat though about the updates to the readme.md

I wonder if there's a way to section it like 

```
# Card Fields
## Using Card Fields Form (recommeneded)
## Using Card Fields Individually
```

elizabethmv (approved):
Had the same though as Shane about the readme, but non-blocking and otherwise looks good!

src/components/cardFields/PayPalCardFieldsForm.tsx
sebastianfdz: The style block with dangerouslySetInnerHTML makes me uncomfortable too 🥲! Maybe we could wrap the `CVVField` and `ExpiryField` inside full width divs?

sebastianfdz: Or directly make our shared `PayPalCardField` container have full width would also solve it. I think this would probably be the cleaner way.

Eg: `<div ref={containerRef} style={{ width: '100%' }} className={className} />;`

sebastianfdz: Got rid of the style block and wrapped both fields in a full width container. Let me know if you have any thoughts!

wsbrunson (approved):
very cool!

PR #433, opened 2024-04-29
Hotfix/form type correction

mchoun:
JIRA Ticket pending

Small PR to fix type for `PayPalCardFieldsForm` component and typo in README.

Initially, we were passing `inputEvents` and `style` to the `PayPalCardFieldsForm` component so that those props were applied to each individual field. We removed `inputEvents` and `style` from this form since the `PayPalCardFieldsProvider` already does this.


sebastianfdz (approved):
Great catch!

elizabethmv (approved):

PR #437, opened 2024-05-24
refactor: simplify template literals in types definition for ScriptReducerAction 

ukrainer:
PR to simplify the type for the ScriptReducerAction (PayPalScriptProvider component).

This refactoring allows the use of this npm package in projects with lower TS versions, which doesn't support template literals in type definitions. Methods such as excluding in tsconfig, skipLibCheck, ts ignore, and no-check were unsuccessful.

The type definition has been simplified, and the test case updated.


gregjopa (approved):
Thanks for contributing @ukrainer!

wsbrunson (approved):

PR #442, opened 2025-02-05
Update payPalCardFieldsTypes.ts

devpato:
Added the fix where CardFields do not recognize the key inside of styles for background and height so when we add this properties now the changes do reflect on the UI


wsbrunson:
@devpato could you please run `npm run changeset:add` and push up the result to this PR?

imbrian:
@devpato thanks for making this change, but the React SDK was moved out of this repo last year. Would you be able to make this change [over here](https://github.com/paypal/paypal-js/blob/main/packages/react-paypal-js/src/types/payPalCardFieldsTypes.ts) instead?

